PR #1, opened 2010-09-01
Making the Shotgun objects inherit from object

schworer:
If anyone is using less than Python 2.2, this patch will break them. However, inheriting from object allows us to use "new style class" features, like property getters and setters. Here's some more info:
http://docs.python.org/release/2.5.2/ref/node33.html

Also this gives me an excuse to try out the new pull request features.

--alex

### DIFF
--- a/shotgun_api3.py
+++ b/shotgun_api3.py
@@ -153,7 +153,7 @@ class Shotgun(object):
 
     def __init__(self, base_url, script_name, api_key, convert_datetimes_to_utc=True, http_proxy=None):
         """
-z       Initialize Shotgun.
+        Initialize Shotgun.
         """
         self.server = None
         if base_url.split("/")[0] not in ("http:","https:"):


Imran-imtiaz48 (commented):
In reviewing this `__init__` method, I noticed a potential improvement in the way URL validation is handled. The current approach checks `base_url.split("/")[0]` against `"http:"` and `"https:"`, which can lead to incorrect results if the URL contains unexpected formatting or additional whitespace. A more robust approach would be to use Python’s `urllib.parse` module to properly parse and validate the scheme, ensuring the code works reliably for different URL formats. Additionally, it would be beneficial to add more descriptive docstrings explaining each parameter, especially `convert_datetimes_to_utc` and `http_proxy`, so that future maintainers understand their purpose and default behavior. Finally, initializing `self.server` to `None` is fine, but it might be worth documenting why it is deferred rather than set during initialization, to avoid confusion for other developers working with this class.

PR #76, opened 2015-01-30
Adds per-project visibility querying on schema methods

francoisjacques:


shotgun_api3/shotgun.py
```diff
@@ -945,23 +945,39 @@ def followers(self, entity):
         
         return self._call_rpc('followers', params)
 
-    def schema_entity_read(self):
+    def schema_entity_read(self, project_entity=None):
```
manneohrstrom: Hey there! It would be great if we could have a `:param: style docstring for this new parameter - just for consistency with other methods! Thanks!

francoisjacques: ya, agree, will do

shotgun_api3/shotgun.py
```diff
@@ -945,23 +945,39 @@ def followers(self, entity):
...
         """Gets all active entities defined in the schema.
 
         :returns: dict of Entity Type to dict containing the display name.
         """
 
-        return self._call_rpc("schema_entity_read", None)
+        params = {}
+
+        if project_entity:
+                params["project"] = project_entity
```
manneohrstrom: Just a heads up that in the github diff, this looks like it's may be indented inconsistently.

shotgun_api3/shotgun.py
manneohrstrom: cheers!

manneohrstrom:
Thanks for the fixes! Looks sweet! :cake:

PR #77, opened 2015-02-10
Ability to instantiate the API via a session id

manneohrstrom:
Adds a constructor parameter `session_token` which makes it possible to create an API instance without either username/password or script/script key. 
A session token can be generated via the new `generate_session_token()` method.

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -1783,10 +1783,12 @@ class Shotgun(object):
 
         :raises ShotgunError: If the server response contains an exception.
         """
+ 
+        ERR_AUTH = 102 # error code for authentication related problems
 
         if isinstance(sg_response, dict) and sg_response.get("exception"):
             
-            if sg_response.get("error_code") == 102:
+            if sg_response.get("error_code") == ERR_AUTH:
                 raise AuthenticationFault(sg_response.get("message", "Unknown Authentication Error"))
 
             else:
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -381,13 +381,6 @@ class Shotgun(object):
         # call to server will only be made once and will raise error
         if connect:
             self.server_caps
-            
-        # if the connection has been established using a username/password combination,
-        # retrieve a session token from the server and then clear the password 
-        # member variable for security reasons
-        if self.config.user_password:
-            self._get_session_token()
-            self.config.user_password = None
 
     # ========================================================================
     # API Functions
@@ -1433,13 +1426,11 @@ class Shotgun(object):
         return url
 
     def authenticate_human_user(self, user_login, user_password):
-        """
-        Authenticate Shotgun HumanUser. HumanUser must be an active account.
-        
-        :param user_login: Login name of Shotgun HumanUser
-        :param user_password: Password for Shotgun HumanUser
-        :return: Dictionary of HumanUser including ID if authenticated, None is unauthorized.
-        """
+        '''Authenticate Shotgun HumanUser. HumanUser must be an active account.
+        @param user_login: Login name of Shotgun HumanUser
+        @param user_password: Password for Shotgun HumanUser
+        @return: Dictionary of HumanUser including ID if authenticated, None is unauthorized.
+        '''
         if not user_login:
             raise ValueError('Please supply a username to authenticate.')
 
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -1500,8 +1500,8 @@ class Shotgun(object):
     def get_session_token(self):
         """
         Get the session token associated with the current session.
-        If a session token has already been established, the token for this 
-        is returned, otherwise a new one is generated on the server and returned.
+        If a session token has already been established, this is returned, 
+        otherwise a new one is generated on the server and returned.
         
         :returns: String containing a session token
         """
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -93,6 +93,10 @@ class Fault(ShotgunError):
     """Exception when server side exception detected."""
     pass
 
+class SessionAuthenticationFault(Fault):
+    """Exception when the server side reports that a session is no longer valid"""
+    pass
+
 # ----------------------------------------------------------------------------
 # API
 
@@ -1607,7 +1611,8 @@ class Shotgun(object):
         # Authenticate using session_id
         elif self.config.session_token:
             auth_params = {
-                "session_token" : str(self.config.session_token)
+                "session_token" : str(self.config.session_token),
+                "reject_if_expired": False
             }
 
         else:
@@ -1789,8 +1794,24 @@ class Shotgun(object):
         """
 
         if isinstance(sg_response, dict) and sg_response.get("exception"):
-            raise Fault(sg_response.get("message",
-                "Unknown Error"))
+            
+            # handle the following cases explicitly
+            #
+            # {'exception': True, 
+            #  'error_code': 102, 
+            #  'message': "Can't authenticate session token 'xxxxxxxx'"}
+            #
+            # {'exception': True, 
+            #  'error_code': 102, 
+            #  'message': "Session token 'xxxxxxxx has expired'"}
+            
+            msg = sg_response.get("message")
+            
+            if msg and ("Can't authenticate session token" in msg or "has expired" in msg):
+                raise SessionAuthenticationFault(msg)
+            else:
+                # raise general Fault            
+                raise Fault(sg_response.get("message", "Unknown Error"))
         return
 
     def _visit_data(self, data, visitor):
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -381,6 +381,13 @@ class Shotgun(object):
         # call to server will only be made once and will raise error
         if connect:
             self.server_caps
+            
+        # if the connection has been established using a username/password combination,
+        # retrieve a session token from the server and then clear the password 
+        # member variable for security reasons
+        if self.config.user_password:
+            self._get_session_token()
+            self.config.user_password = None
 
     # ========================================================================
     # API Functions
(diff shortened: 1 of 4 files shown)


shotgun_api3/shotgun.py
```diff
@@ -308,9 +327,11 @@ def __init__(self,
         self.config.script_name = script_name
         self.config.user_login = login
         self.config.user_password = password
+        self.config.session_token = session_token
         self.config.sudo_as_login = sudo_as_login
         self.config.convert_datetimes_to_utc = convert_datetimes_to_utc
         self.config.no_ssl_validation = NO_SSL_VALIDATION
+        self.config.raw_http_proxy = http_proxy
```
robblau: This could use a comment as to how it could be used.  Since it isn't used in the API itself it isn't obvious.

shotgun_api3/shotgun.py
```diff
@@ -298,9 +317,9 @@ def __init__(self,
                 raise ValueError("script_name provided without api_key")
 
         # Can't use 'all' with python 2.4
-        if len([x for x in [script_name, api_key, login, password] if x]) == 0:
+        if len([x for x in [session_token, script_name, api_key, login, password] if x]) == 0:
             if connect:
-                raise ValueError("must provide either login/password "
+                raise ValueError("must provide either login/password, session_token "
```
robblau: Should strip out "either" since there are now more than two options.

shotgun_api3/shotgun.py
```diff
@@ -1473,20 +1494,33 @@ def update_project_last_accessed(self, project, user=None):
         record = self._call_rpc("update_project_last_accessed_by_current_user", params)
         result = self._parse_records(record)[0]
 
-
-    def _get_session_token(self):
-        """Hack to authenticate in order to download protected content
-        like Attachments
+    def generate_session_token(self):
```
robblau: Is there a reason to call this generate_session_token rather than matching the server side call and calling it "get_session_token".  Get feels like it implies a Shotgun round trip where generate doesn't to me.

shotgun_api3/shotgun.py
```diff
@@ -1473,20 +1494,33 @@ def update_project_last_accessed(self, project, user=None):
...
+        """
+        Generate a session token that can be used to authenticate the currently 
+        logged in user or script.
+        
+        :returns: String containing a session token
         """
-        if self.config.session_token:
-            return self.config.session_token
-
```
robblau: Also, why not return the cached session if it is set?  The reason for the difference between this and the protected _get_session_token is not clear.

shotgun_api3/shotgun.py
```diff
@@ -1573,6 +1608,13 @@ def _auth_params(self):
                 "script_key" : str(self.config.api_key),
             }
 
+        # Authenticate using session_id
+        elif self.config.session_token:
+            auth_params = {
+                "session_token" : str(self.config.session_token),
+                "reject_if_expired": False
```
robblau: Why is this False?  Shouldn't authentication fail if the session is expired?

shotgun_api3/shotgun.py
```diff
@@ -1752,8 +1794,24 @@ def _response_errors(self, sg_response):
...
+            #  'message': "Can't authenticate session token 'xxxxxxxx'"}
+            #
+            # {'exception': True, 
+            #  'error_code': 102, 
+            #  'message': "Session token 'xxxxxxxx has expired'"}
+            
+            msg = sg_response.get("message")
+            
+            if msg and ("Can't authenticate session token" in msg or "has expired" in msg):
+                raise SessionAuthenticationFault(msg)
```
robblau: Relying on the error message does feel a little iffy, it would be pretty easy for a change in the Shotgun codebase to break this and we wouldn't necessarily realize it.  At the very least there should be a comment in the Shotgun code that the python API depends on the wording of this errror.

shotgun_api3/shotgun.py
manneohrstrom: heh, initially i had a comment but removed it given the coding style. It's back now! :)

shotgun_api3/shotgun.py
manneohrstrom: ok. will fix!

shotgun_api3/shotgun.py
manneohrstrom: hehe, i was actually using `generate` to try to emphasize that it is doing something on the server - not just getting some existing data, but in fact fabricating a new record. `get_session_token` sounded more to me like you are getting an existing token - the idea with generate was to imply that a new one is being returned. (confusingly, also noticed that the there is a `get_attachment_download_url()` method as the only other "getter" in the shotgun API and that isn't actually talking to the server!).

Do you think it's important to match the server side call? I thought this was more of a private detail?

I don't have a strong opinion at all, get is totally fine too, let me know if you think I should change it!

shotgun_api3/shotgun.py
manneohrstrom: The reason was that generate_session_token *generates* a new session token each time it is called. However, we may not need this as much now - it turns out the sessions don't expire after a certain time, they expire after a certain time of *inactivity*, making it (I think) pointless to keep creating new session tokens. As long as you keep using the token you have, you should be fine!

...which also means that we can probably return the cached token rather than making a new, and subsequently change it from generate_to_get() ! :)

shotgun_api3/shotgun.py
manneohrstrom: Oops, some leftovers from testing. Thanks for spotting this. Shouldn't be in there.

shotgun_api3/shotgun.py
manneohrstrom: good point! I'll make sure that gets added.

manneohrstrom: The errors have an error code as well - i think the problem here is that *different* error messages coming from the server are using the *same* error code - where probably each error should have been using a different error code (at least if we wanted to use the codes for raising better exceptions on the client side). I am guessing it is too late to change on the server side too - that may risk breaking something else... Or do you think there is some good way to handle both scenarios?

robblau: Just looked at the shotgun code and 102 is ERR_AUTH, so we can be sure that it is a failed authentication.  Is that enough for the use case you have (ie if you get a 102 and were trying to use a session token then you know what happened)?

Other than that... even the vagueness of the check seems bad "has expired" could match something else in the future.

And looking through the possible session errors, there are more:
User associated with session token '#{auth_args['session_token']} no longer authorized.'
Corrupted session token '#{auth_args['session_token']}'

and another that wouldn't be 102 on poorly formatted session token:
API #{mtd}() session_uuid is not a valid UUID argument:\n#{API3Handler.ppstr(auth_args)}

Feels like simply if you get a 102 when trying to authenticate with a session then you know what happened and matching the message shouldn't be needed.

manneohrstrom: Well, the ERR_AUTH will be passed back for all authentication issues, where as I was trying to get a special exception raised when a session has expired. Hence the string matching. But maybe it's enough to determine that it's an authentication issue? (e.g. bad username/password/permissions related/sessions etc). As basing it on the error code is safer and nicer, I'll make those changes first and then we can see if we need further refinement later down the line....

shotgun_api3/shotgun.py
```diff
@@ -1752,8 +1785,13 @@ def _response_errors(self, sg_response):
         """
 
         if isinstance(sg_response, dict) and sg_response.get("exception"):
-            raise Fault(sg_response.get("message",
-                "Unknown Error"))
+            
+            if sg_response.get("error_code") == 102:
+                raise AuthenticationFault(sg_response.get("message", "Unknown Authentication Error"))
```
robblau: That feels less fragile.  Only thing I'd say is to comment on that 102 to say where it comes from.  Otherwise it is just a magic number.

shotgun_api3/shotgun.py
```diff
@@ -1473,10 +1497,13 @@ def update_project_last_accessed(self, project, user=None):
...
         result = self._parse_records(record)[0]
 
-
-    def _get_session_token(self):
-        """Hack to authenticate in order to download protected content
-        like Attachments
+    def get_session_token(self):
+        """
+        Get the session token associated with the current session.
+        If a session token has already been established, the token for this
```
jfboismenu-adsk: "the token for this is returned".
Sounds odd. What is this "this" you are speaking of?
"it is returned" feels better IMHO.

shotgun_api3/shotgun.py
```diff
@@ -1473,10 +1497,13 @@ def update_project_last_accessed(self, project, user=None):
         record = self._call_rpc("update_project_last_accessed_by_current_user", params)
         result = self._parse_records(record)[0]
 
-
-    def _get_session_token(self):
-        """Hack to authenticate in order to download protected content
-        like Attachments
+    def get_session_token(self):
```
jfboismenu-adsk: The original comments dissapeared, but method is still important even tough we are not constantly refreshing the token anymore in Toolkit because when you login using username/password then you need to be able to generate the token that first time.

jfboismenu-adsk:
You reverted your commit where you cleared the username/password after authenticating. Why? Is it just performance? Or it is because it's not part of our current mandate?

shotgun_api3/shotgun.py
manneohrstrom: will fix!

shotgun_api3/shotgun.py
manneohrstrom: fixed!

shotgun_api3/shotgun.py
manneohrstrom: This should still work - i just made the method public `get_session_token` rather than `_get_session_token` - so you can log in with username and password, then run `get_session_token()` to get a session token back. Subsequent calls to `get_session_token` will then return the same (cached) value over and over again - which should be fine given  that the session doesn't expire as long as it is being used. 

Does that sound right to you?

manneohrstrom:
I reverted the commit about the username/password because there is a subtle regression which makes it trickier to deal with than I first thought: If you have a shotgun session which was instantiated with a login/password, this session is currently operational **forever**, because it keeps sending the username and password back to the server. It is, however, storing the password as a member variable, which is not good practice. 

If we dropped the password at construction time and instead continued with a session token, the API session wouldn't be valid forever anymore -- if you left it idle for 24 hours and then picked it back up again it would raise an `AuthenticationFault`. We have some places where we have long running loops and stuff going, so although subtle, we probably need to communicate this change to clients explicitly prior to a release. (Where as the rest of the stuff in this pull request is just adding new features!)

shotgun_api3/shotgun.py
jfboismenu-adsk: Yes, that's perfectly fine. I simply added my comment to state that it was still important to keep it in, since previous comments on that method indicated that you were not sure it would still be needed.

jfboismenu-adsk:
Gotcha. Totally makes sense.

manneohrstrom:
this is now merged into the toolkit security branch.

PR #79, opened 2015-03-13
Ticket/28441 hide template projects by default

nemoDreamer:
We're adding `Project.is_template` in v6.0.0 to support multiple template projects. Unlike `include_archived_projects` (which defaults to `True`), the new `include_template_projects` CRUD flag defaults to `False`.
- I did some refactoring as I felt dirty copy/pasting `ensure_include_archived_projects` 
- The rest is IF statements copied from existing `include_archived_flag` support w/ the logic reversed
- Moved the flag tests into their own file, to avoid needing to run the gargantuan `test_api.py`

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -171,6 +171,13 @@ class ServerCapabilities(object):
             'label': 'include_archived_projects parameter'
         })
 
+    def ensure_include_template_projects(self):
+        """Wrapper for ensure_support"""
+        self._ensure_support({
+            'version': (6, 0, 0),
+            'label': 'include_template_projects parameter'
+        })
+
     def ensure_per_project_customization(self):
         """Wrapper for ensure_support"""
         return self._ensure_support({
@@ -485,7 +492,8 @@ class Shotgun(object):
         return self._call_rpc("info", None, include_auth_params=False)
 
     def find_one(self, entity_type, filters, fields=None, order=None,
-        filter_operator=None, retired_only=False, include_archived_projects=True):
+        filter_operator=None, retired_only=False,
+        include_archived_projects=True, include_template_projects=False):
         """Calls the find() method and returns the first result, or None.
 
         :param entity_type: Required, entity type (string) to find.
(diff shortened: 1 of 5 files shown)


nemoDreamer:
Questions: why is the test-suite so insanely slow locally? The Travis test passed much faster. Is it something to do w/ us hitting our puny VM? Does `setUp` run before every test, or only once per test class? Because the majority of these don't need a full DB reset.

shotgun_api3/shotgun.py
```diff
@@ -2102,4 +2152,6 @@ def _translate_filters_simple(sg_filter):
 
     return condition
 
-     
+def _version_str(version):
+    """Converts a tuple of int's to a '.' separated str"""
+    return '.'.join(map(str, version))
```
nemoDreamer: Can't believe that Python's `join` is called on the separator... It should be a functionality of an enumerable, not a string... Weird...!

nemoDreamer:
Ran the entire suite on 5.4, then 6.0, then 6.0-28441, just to make sure all was coming out green. But took aaaaaaages :laughing:

nemoDreamer:
*Note:* see shotgunsoftware/shotgun#1119 for CRUD changes.

nemoDreamer:
CRUD does not currently seem to handle the new flag correctly when coming from an API call, so tests are failing "correctly".

We're putting this on the back-burner, since at least like this, we're not breaking existing API functionality (running the old API agains a new site comes up green).

shotgun_api3/shotgun.py
```diff
@@ -134,26 +134,53 @@ def __init__(self, host, meta):
...
+    def _ensure_support(self, feature, raise_hell=True):
+        """Checks the server version supports a given feature, raises an
         exception if it does not.
 
-        :raises ShotgunError: The current server version does not support json
-        """
-        if not self.version or self.version < (2, 4, 0):
-            raise ShotgunError("JSON API requires server version 2.4 or "\
-                "higher, server is %s" % (self.version,))
+        :param feature: dict supported version and human label { 'version': (int, int, int), 'label': str }
```
jfboismenu-adsk: Please update the documentation.

shotgun_api3/shotgun.py
```diff
@@ -134,26 +134,53 @@ def __init__(self, host, meta):
...
+
+    def ensure_json_supported(self):
+        """Wrapper for ensure_support"""
+        return self._ensure_support({
+            'version': (2, 4, 0),
+            'label': 'JSON API'
+        })
+
+    def ensure_include_archived_projects(self, value=True):
+        """Wrapper for ensure_support"""
```
jfboismenu-adsk: value not document.

shotgun_api3/shotgun.py
```diff
@@ -134,26 +134,53 @@ def __init__(self, host, meta):
...
+        # This defaults to True on the server
+        # So we only need to raise a version error if it's False
+        return self._ensure_support({
+            'version': (5, 3, 14),
+            'label': 'include_archived_projects parameter'
+        }, (value == False))
+
+    def ensure_include_template_projects(self, value=False):
+        """Wrapper for ensure_support"""
```
jfboismenu-adsk: value not documented.

manneohrstrom:
@nemoDreamer can this be closed?

nemoDreamer:
No idea, @manneohrstrom : it's 3 years old, and I have no idea if it's still wanted.

patrick-hubert-adsk:
Dummy comment to keep my user active

PR #80, opened 2015-03-16
#28697 Add environment variable for specifying location of certs file

kporangehat:
Adds a check for the `SHOTGUN_API_CACERTS` environment variable for the location of the certificates file to use when validating SSL connections. Allows studios to use their own certificate store automatically without having to instantiate the Shotgun API instance with this parameter explicitly. This is especially useful when using distributed applications like SG Toolkit with a server that has an internal or corporate CA signed certificate. Also updates the bundled `cacerts.txt` file to the latest version published by Mozilla (via certifi.io)

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -314,7 +314,10 @@ class Shotgun(object):
         self.config.convert_datetimes_to_utc = convert_datetimes_to_utc
         self.config.no_ssl_validation = NO_SSL_VALIDATION
         self._connection = None
-        self.__ca_certs = ca_certs or os.environ.get('SHOTGUN_API_CACERTS')
+        if ca_certs is not None:
+            self.__ca_certs = ca_certs
+        else:
+            self.__ca_certs = os.environ.get('SHOTGUN_API_CACERTS')
 
         self.base_url = (base_url or "").lower()
         self.config.scheme, self.config.server, api_base, _, _ = \
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -265,7 +265,9 @@ class Shotgun(object):
         :param connect: If True, connect to the server. Only used for testing.
 		
 		:param ca_certs: The path to the SSL certificate file. Useful for users
-		who would like to package their application into an executable.
+		who would like to package their application into an executable or for
+        corporate or internal CAs. Note that providing this parameter explicitly 
+        will override any value in the SHOTGUN_API_CACERTS environment variable. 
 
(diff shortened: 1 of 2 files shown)


shotgun_api3/shotgun.py
```diff
@@ -312,7 +314,7 @@ def __init__(self,
         self.config.convert_datetimes_to_utc = convert_datetimes_to_utc
         self.config.no_ssl_validation = NO_SSL_VALIDATION
         self._connection = None
-        self.__ca_certs = ca_certs
+        self.__ca_certs = ca_certs or os.environ.get('SHOTGUN_API_CACERTS')
```
robblau: Would be nicer to structure this as

```python
if ca_certs is not None:
    self.__ca_certs = ca_certs
else:
   self.__ca_certs = os.environ.get('SHOTGUN_API_CACERTS')
```

That way if the passed in param for ca_certs just happens to have a False value the current behavior will be preserved.  Using that or construct can be more trouble than it is worth.

robblau:
One more thought... do we have any unit test for how cacert can be specified? That should be updated for the new env var.

kporangehat:
We don't have any tests currently. Not sure what we'd add...?

shotgun_api3/shotgun.py
```diff
@@ -265,7 +265,9 @@ def __init__(self,
         :param connect: If True, connect to the server. Only used for testing.
 		
 		:param ca_certs: The path to the SSL certificate file. Useful for users
-		who would like to package their application into an executable.
+		who would like to package their application into an executable or for
+        corporate or internal CAs. Note that providing this parameter explicitly 
+        will override any value in the SHOTGUN_API_CACERTS environment variable.
```
manneohrstrom: We are referring to the `SHOTGUN_API_CACERTS` environment variable here as if it is a known thing, but it hasn't yet been mentioned in the docstring (as far as I can tell). Also, the previous doc string isn't great - "The path to the SSL certificate file...." - "**the** SSL certicate" seems to imply that this is required where it is optional and an expert setting. 

Suggested doc string to make this clearer:

```
:param ca_certs: Optional path to an external SSL certificate file. By default, the Shotgun API 
will use its own built-in certificates file, but if you are running a corporate or internal CA, 
it may be necessary to override these defaults. You can override it either by passing 
it in via this parameter or by setting the environment variable `SHOTGUN_API_CACERTS`.
In the case both are set, this parameter will take precedence.
```

KP, you are the master of grammar and words, so I am sure you will tweak and correct this further. Hope it's useful.

PR #83, opened 2015-03-25
#28822 ensure we don't try and sort by pivot_column fields by accident

kporangehat:
`pivot_column` fields aren't sortable so in the case where we're sorting by an arbitrary field, ensure it's not a `pivot_column`. Iterate though the field list until we reach one that works.


manneohrstrom:
looks good!

kporangehat:
resolved by a64c55ba2fc3613ad8b062c2a81635df23028744

PR #85, opened 2015-05-13
For #29803: Added support for 2FA

robblau:
Added support for 2 factor authentication (2FA) to the API so it is ready when Shotgun supports this. 
The "auth_token" parameter added to the authentication process defaults to "None" and can be ignored by all clients until 2FA is made available.


shotgun_api3/shotgun.py
```diff
@@ -97,6 +97,10 @@ class AuthenticationFault(Fault):
     """Exception when the server side reports an error related to authentication"""
     pass
 
+class MultiFactorAuthenticationFault(Fault):
+    """Exception when the server side reports an error related to missing 2FA credentials"""
```
robblau: This is raised on missing 2FA credentials, right?

Seems like there are some new possible error conditions... do we know what error is raised if:
- 2FA is turned on and the password is wrong but a token is specified?
- 2FA is turned on and the password is right but a wrong token is specified?
- 2FA is turned off and the token is specified?

If the error is only when the 2FA token is needed but missing, then maybe a name like "MissingMultiFactorAuthenticationFault" would be less ambiguous (though longer)?

Also, we should watch out for line length here.

shotgun_api3/shotgun.py
```diff
@@ -321,6 +327,10 @@ def __init__(self,
         the server. If password is provided, then login must be as well and
         neither script_name nor api_key can be provided.
         
+        :param auth_token: The one time token required to authenticate to
+        a server with 2FA turned on. If auth_token is provided, then login and password
+        must be as well and neither script_name nor api_key can be provided.
```
robblau: There isn't a check for script_name + api_key + auth_token by my read.

A call that specifies all three looks like it wouldn't raise an error but would just ignore the auth_token.  Either that should be fixed or this docstring should be updated.

shotgun_api3/shotgun.py
```diff
@@ -1519,7 +1534,7 @@ def get_attachment_download_url(self, attachment):
                 None, None, None))
         return url
 
-    def authenticate_human_user(self, user_login, user_password):
+    def authenticate_human_user(self, user_login, user_password, auth_token):
```
robblau: Requiring an auth_token here is a non-backward compatible change that could break client scripts.  It should have a default value so existing scripts can be used without change.

shotgun_api3/shotgun.py
```diff
@@ -278,6 +283,7 @@ def __init__(self,
                  ca_certs=None,
                  login=None,
                  password=None,
+                 auth_token=None,
```
robblau: New optional parameters should be added at the end in case anybody is using the constructor passing in arguments by position.  Otherwise it is a non-backward compatible change.

shotgun_api3/shotgun.py
```diff
@@ -1867,14 +1888,15 @@ def _response_errors(self, sg_response):
...
         ERR_AUTH = 102 # error code for authentication related problems
+        ERR_2FA  = 106 # error code when 2FA authentication is required but no 23FA token provided.
 
         if isinstance(sg_response, dict) and sg_response.get("exception"):
-            
             if sg_response.get("error_code") == ERR_AUTH:
                 raise AuthenticationFault(sg_response.get("message", "Unknown Authentication Error"))
-
+            elif sg_response.get("error_code") == ERR_2FA:
+                raise MultiFactorAuthenticationFault(sg_response.get("message", "Unknown 2FA Authentication Error"))
```
robblau: The Authentication error is unknown because it is a generic catch all error for any of various authentication problems.  Is there a better error message possible for this error code?  Also, is there a reason to use 2FA here rather than Multifactor?  The wording in the error code variable/fault and the error message seems inconsistent.

shotgun_api3/shotgun.py
manneohrstrom: Please also add the parameter to the doc string and ideally also information about which exceptions you can except to get back in various situations (wrong auth token, missing auth token, etc).

shotgun_api3/shotgun.py
daigles: Changed to "Missing...Fault", The "auth_token" is ignored (even if given) when 2FA is not turned on.

shotgun_api3/shotgun.py
daigles: added the test case...

shotgun_api3/shotgun.py
daigles: Oops. Fixed and documented.

shotgun_api3/shotgun.py
daigles: done.

shotgun_api3/shotgun.py
daigles: The "Unknown error" is the default value returned if there is no message field in the response. I'll change MultiFactor... to TwoFactor since this is what we have server side.

PR #87, opened 2015-07-10
#29814 Added mockgun mocker to Shotgun API

manneohrstrom:
This adds a highly experimental 'Mockgun' Shotgun API mocker to the Shotgun API. This is being moved across from the Toolkit unit test rig into the Shotgun API to allow for a wider range of use and reuse. 

Mockgun is highly experimental at this point. It only supports a subset of the Shotgun API and we are likely to make big and potentally breaking changes to this in the future.


nemoDreamer:
Love the name, @manneohrstrom ! :candy:

PR #90, opened 2015-09-03
#30015 New API methods for search, activity stream and note thread access

manneohrstrom:
Adds three new methods:

- `text_search()` gives access to the shotgun global search and auto completer.
- `activity_stream_read()` gives access to the shotgun actitivty stream.
- `note_thread_read()` gives access to a note thread, including replies and attachments, via a single call.

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -1627,7 +1627,9 @@ class Shotgun(object):
 
     def note_thread(self, note_id, entity_fields=None):
         """
-        Returns the full conversation for a note, including replies and attachments:
+        Returns the full conversation for a given note, including 
+        replies and attachments. Returns a complex data structure
+        on the following form:
         
             [{'content': 'Please add more awesomeness to the color grading.',
               'created_at': '2015-07-14 21:33:28 UTC',
@@ -1656,28 +1658,29 @@ class Shotgun(object):
                        'type': 'HumanUser',
                        'valid': 'valid'}}]
 
-        The list is returned in descneding chronological order.
+        The list is returned in descending chronological order.
         
         If you wish to include additional fields beyond the ones that are 
         returned by default, you can specify these in an entity_fields 
         dictionary. This dictonary should be keyed by entity type and each
         key should contain a list of fields to retrieve, for example:
         
-            { "Note": ["created_by.HumanUser.image", 
-                      "addressings_to", 
-                      "playlist", 
-                      "user" ],
-              "Reply": ["content"], 
+            { "Note":       ["created_by.HumanUser.image", 
+                             "addressings_to", 
+                             "playlist", 
+                             "user" ],
+              "Reply":      ["content"], 
               "Attachment": ["filmstrip_image", 
                             "local_storage", 
                             "this_file", 
                             "image"]
             }
         
-        :param note_id: The id for the note to be retrieved
-        :param entity_fields: Additional fields to retrieve as part of the request. 
-                              See above for details.
-        :returns: Complex data structure 
+        :param note_id: The entity id for the note to be retrieved
+        :param entity_fields: Additional fields to retrieve as part 
+                              of the request. See above for details.
+                              
+        :returns: A list of dictionaries. See above for example. 
         """        
 
         if self.server_caps.version and self.server_caps.version < (6, 2, 0):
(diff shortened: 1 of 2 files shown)


shotgun_api3/shotgun.py
```diff
@@ -1623,6 +1622,276 @@ def update_project_last_accessed(self, project, user=None):
         record = self._call_rpc("update_project_last_accessed_by_current_user", params)
         result = self._parse_records(record)[0]
 
+
+
+    def note_thread_read(self, note_id, entity_fields=None):
+        """Returns the full conversation for a given note, including 
+        replies and attachments.
+        
+        Returns a complex data structure on the following form:
```
robblau: Should read:
```
Returns a complex data structure in the following form
```

shotgun_api3/shotgun.py
```diff
@@ -1623,6 +1622,276 @@ def update_project_last_accessed(self, project, user=None):
...
+        
+        params = { "note_id": note_id, "entity_fields": entity_fields }
+
+        record = self._call_rpc("note_thread_contents", params)
+        result = self._parse_records(record)
+        return result
+
+
+    def text_search(self, text, entity_types, project_ids=None, limit=None):
+        """
```
robblau: Comment should start on the same line as the triple quote for consistency.

shotgun_api3/shotgun.py
```diff
@@ -1623,6 +1622,276 @@ def update_project_last_accessed(self, project, user=None):
...
+        Searches across selected entity types for a given text.
```
robblau: Should be more like

```
Searches across the specified entity types for the given text.
```

shotgun_api3/shotgun.py
```diff
@@ -1623,6 +1622,276 @@ def update_project_last_accessed(self, project, user=None):
...
+                                  ["sg_status_list", "is", "ip"]] } }
+          
+          The general syntax for this data structure is:
+          { entity_type: { filter_operator: all|any,
+                           filters: shotgun_filter_list } }
+            
+          The filter_operator key is optional and if not specified, 'all' 
+          will be used.
+        
+        A dictionary with keys terms and matches will be returned:
```
robblau:
```
with the keys 'terms' and 'matches' will be returned:
```

shotgun_api3/shotgun.py
```diff
@@ -1623,6 +1622,276 @@ def update_project_last_accessed(self, project, user=None):
...
+        for (entity_type, filter_dict) in entity_types.iteritems():
+            
+            if not isinstance(filter_dict, dict):
+                raise ValueError("value of entity_types['%s'] must "
+                                 "be a dictionary" % entity_type)
+            
+            if "filters" in filter_dict:
+                # there is a filter query inside the dict. 
+                # convert it into native form 
+                filter_operator = filter_dict.get("filter_operator")
```
robblau: Is there a reason to make this specific to the dictionary type filter syntax?  Seems like it would be easier to explain and document if we just accept (per entity type key) the same structures we support for general filter arguments... ie lists in the form [["name", "contains", "foo"]] or the dictionary syntax.  Then we can just update the docs to refer to the general page on how to filter.

shotgun_api3/shotgun.py
```diff
@@ -1623,6 +1622,276 @@ def update_project_last_accessed(self, project, user=None):
...
+                   "max_results": limit }
+
+        record = self._call_rpc("query_display_name_cache", params)
+        result = self._parse_records(record)[0]
+        return result
+
+
+    def activity_stream_read(self, entity_type, entity_id, entity_fields=None, 
+                             min_id=None, max_id=None, limit=None):
+        """
```
robblau: Comment should start on the same line for consistency.

shotgun_api3/shotgun.py
```diff
@@ -1623,6 +1622,276 @@ def update_project_last_accessed(self, project, user=None):
...
+        Retrieves activity stream data from Shotgun.
+        This data corresponds to the data that is displayed on the 
+        Activity tab in the Shotgun Web UI.
```
robblau:
```
Activity tab for an entity in the Shotgun Web UI.
```

Just to make sure people don't confuse it with the Inbox.

shotgun_api3/shotgun.py
```diff
@@ -1623,6 +1622,276 @@ def update_project_last_accessed(self, project, user=None):
...
+        ]}
+        
+        The main payload of the return data can be found inside the 'updates' 
+        key, containing a list of dictionaries. This list is always returned 
+        in descending date order. Each item may contain different fields 
+        depending on their update type. The primary_entity key represents the 
+        main Shotgun entity that is associated with the update. By default, 
+        this entity is returned with a set of standard fields. By using the 
+        entity_fields parameter, you can extend the returned data to include 
+        additional fields. If you for example wanted to return the asset type
```
robblau:
```
If for example you wanted to
```

shotgun_api3/shotgun.py
manneohrstrom: Good point! I'll investigate this.

PR #92, opened 2015-10-23
32998 sanitize auth dicts

pboucher:


RobertLBelanger:
There is always a risk that a new auth param be added without update _sanitize_auth_params.

I guess, this could go in our security best practices/recipes.

kporangehat:
Looks good. You could argue we don't really even need to show the params in that exception, but it's a nice convenience and sanitation looks good.

PR #95, opened 2015-11-13
Fix garbage collection for Shotgun instances

bchapman:
Found that Shotgun instances weren't getting garbage collected.
Did some research and found that it's due to the this line in __int__

```
if ensure_ascii:
    self._json_loads = self._json_loads_ascii
```

This causes python not to garbage collect any Shotgun instances

Example:
```
>>> import gc
>>> sg = shotgun_api3.Shotgun(**params)
>>> gc.get_referrers(sg)
[<bound method Shotgun._json_loads_ascii of <trak.shotgun.Shotgun object at 0x10542dd10>>, ...]
```

Changed it to store the `ensure_ascii` as variable on the class and check in `_json_loads` which method to use.

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -446,8 +446,7 @@ class Shotgun(object):
             proxy_addr = "http://%s%s:%d" % (auth_string, self.config.proxy_server, self.config.proxy_port)
             self.config.proxy_handler = urllib2.ProxyHandler({self.config.scheme : proxy_addr})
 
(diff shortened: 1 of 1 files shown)


kporangehat:
Hey Brennan, 

Thanks for the insight and the PR! We were mostly offline last week at our yearly summit so apologies for not chiming in sooner. This change does look pretty innocuous. We'll assign someone to run the tests against it and try and bring it in. Will follow up shortly.

PR #96, opened 2016-01-08
#34265 add support for SHOTGUN_DISABLE_SSL_VALIDATION env variable

kporangehat:
We are updating our server certificates to more secure ones signed with SHA-2. Older versions of Python may have some issues with this change so we're tying to help accommodate these users by making it a little easier to disable SSL certificate verification without having to modify the actual Python API code, and using an environment variable instead.

The impact of disabling validation is that the identity of the Shotgun server can't be determined for sure. The communication is still encrypted, but this leaves the client subject to a man-in-the-middle attack. Such attacks are hard to perpetrate, but remain a possibility


manneohrstrom:
With the above code, I would get the following logic in a bash shell:
```
export SHOTGUN_DISABLE_SSL_VALIDATION=1 --> validation will be disabled
export SHOTGUN_DISABLE_SSL_VALIDATION=0 --> validation will also be disabled
export SHOTGUN_DISABLE_SSL_VALIDATION   --> no effect
```

Is this perhaps confusing? If we did something like this:

```
if os.environ.get("SHOTGUN_DISABLE_SSL_VALIDATION", "0") != "0"
```

, `export SHOTGUN_DISABLE_SSL_VALIDATION=0` would *not* trigger the disabled state, which i feel is more what I would expect. (The above could could probably be done more elegantly, and I haven't tested it on windows). Does that make sense?

Also, I think it would be nice if we could add unit tests if possible, since environment variables sometimes behave subtly different on different platforms. Also, is this documented anywhere? I did a quick check through the docs, but couldn't seem to find any mention of `NO_SSL_VALIDATION`. Apologies if I missed it!

kporangehat:
Agreed. Pulled the release until it's fixed. :zap:

PR #97, opened 2016-01-10
#34265 add handling for Python versions incompatible with SHA-2

kporangehat:
#### Change in behavior
We are updating our server certificates to more secure ones signed with SHA-2. Some older versions of Python will have issues with this change as they do not support SHA-2 encryption. In order to try and prevent scripts from breaking, when the API encounters a version of Python that is incompatible with SHA-2, it will automatically turn off certificate verification and try the request again. If the validation still fails for some reason, the error will be raised, otherwise the request succeeds and validation will remain off for the remaining life of the connection.

This behavior of having certificate validation off, is actually the default in Python versions < v2.7.9. Up to this point we have been electing to enhance the default level of security. Your connection is still encrypted when certificate validation is off, but the server identity cannot be verified.

#### Logging warnings
When the connection falls back to not validating the certificate, a warning message is generated in the logs:

```
Warning: shotgun_api3 : SSLHandshakeError: this Python installation is incompatible with certificates signed with SHA-2. Disabling certificate validation. For more information, see http://blog.shotgunsoftware.com/2016/01/important-ssl-certificate-renewal-and.html
```

#### `SHOTGUN_FORCE_CERTIFICATE_VALIDATION` environment variable support
There is also support for the `SHOTGUN_FORCE_CERTIFICATE_VALIDATION` environment variable which when set (the value does not matter), will prevent disabling certificate verification and will instead raise an exception.

#### SSL info added to user-agent
Adds info showing the OpenSSL version (if available) and whether certificate validation is enabled or not, to the user-agent string: 

`ssl OpenSSL 1.0.2d 9 Jul 2015 (no-validate)` when validation is disabled
`ssl OpenSSL 1.0.2d 9 Jul 2015 (validate)` when validation is enabled
`ssl OpenSSL unknown (validate)` when the ssl version cannot be determined (< Python 2.7)

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -73,9 +73,8 @@ except ImportError, e:
     if "SHOTGUN_FORCE_CERTIFICATE_VALIDATION" in os.environ:
         raise ImportError("%s. SHOTGUN_FORCE_CERTIFICATE_VALIDATION environment variable prevents "
                           "disabling SSL certificate validation." % e)
-    else:
-        LOG.debug("ssl not found, disabling certificate validation")
-        NO_SSL_VALIDATION = True
+    LOG.debug("ssl not found, disabling certificate validation")
+    NO_SSL_VALIDATION = True
 
 # ----------------------------------------------------------------------------
 # Version
@@ -220,20 +219,11 @@ class ClientCapabilities(object):
 
         # extract the OpenSSL version if we can. The version is only available in Python 2.7 and
         # only if we successfully imported ssl
-        ssl_version_str = None
-        ssl_version_num = "unknown"
+        self.ssl_version = "unknown"
         try:
-            ssl_version_str = ssl.OPENSSL_VERSION
+            self.ssl_version = ssl.OPENSSL_VERSION
         except AttributeError, NameError:
             pass
-        if ssl_version_str:
-            try:
-                ssl_version_num = re.search("openssl\s*([\S.]+).*", ssl_version_str, 
-                                            flags=re.IGNORECASE).group(1)
-            except AttributeError:
-                # the version string format isn't right
-                pass
-        self.ssl_version = ssl_version_num
 
     def __str__(self):
         return "ClientCapabilities: platform %s, local_path_field %s, "\
@@ -1291,7 +1281,7 @@ class Shotgun(object):
         
         self._user_agents = ["shotgun-json (%s)" % __version__,
                              "Python %s (%s)" % (self.client_caps.py_version, ua_platform),
-                             "ssl %s %s" % (self.client_caps.ssl_version, validation_str)]
+                             "ssl %s (%s)" % (self.client_caps.ssl_version, validation_str)]
 
 
     def set_session_uuid(self, session_uuid):
@@ -2033,7 +2023,7 @@ class Shotgun(object):
         self.config.no_ssl_validation = True
         NO_SSL_VALIDATION = True
         # reset ssl-validation in user-agents
-        self._user_agents = ["ssl %s no-validate" % self.client_caps.ssl_version 
+        self._user_agents = ["ssl %s (no-validate)" % self.client_caps.ssl_version 
                              if ua.startswith("ssl ") else ua 
                              for ua in self._user_agents] 
 
@@ -2225,10 +2215,11 @@ class Shotgun(object):
                                 "certificates signed with SHA-2. Disabling certificate validation. "
                                 "For more information, see http://blog.shotgunsoftware.com/2016/01/"
                                 "important-ssl-certificate-renewal-and.html")
+                    self._turn_off_ssl_validation()
+                    # reload user agent to reflect that we have turned off ssl validation
+                    req_headers["user-agent"] = "; ".join(self._user_agents)
+                
                 self._close_connection()
-                self._turn_off_ssl_validation()
-                # reload user agent to reflect that we have turned off ssl validation
-                req_headers["user-agent"] = "; ".join(self._user_agents)
                 if attempt == max_rpc_attempts:
                     raise
             except Exception:
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -70,7 +70,7 @@ NO_SSL_VALIDATION = False
 try:
     import ssl        
 except ImportError, e:
-    if os.environ.get("SHOTGUN_FORCE_CERTIFICATE_VALIDATION") not in [0, "0", "", None]:
+    if "SHOTGUN_FORCE_CERTIFICATE_VALIDATION" in os.environ:
         raise ImportError("%s. SHOTGUN_FORCE_CERTIFICATE_VALIDATION environment variable prevents "
                           "disabling SSL certificate validation." % e)
     else:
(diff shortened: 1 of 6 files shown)


shotgun_api3/shotgun.py
```diff
@@ -66,19 +66,20 @@
...
 SG_TIMEZONE = SgTimezone()
 
-
 NO_SSL_VALIDATION = False
 try:
-    import ssl
-    if os.environ.get("SHOTGUN_DISABLE_SSL_VALIDATION", False):
+    import ssl        
+except ImportError, e:
+    if os.environ.get("SHOTGUN_FORCE_CERTIFICATE_VALIDATION") not in [0, "0", "", None]:
```
robblau: This form of checking the value of the env var is feeling pretty ugly.  I think

```python
if "SHOTGUN_FORCE_CERTIFICATE_VALIDATION" in os.environ:
   ...
```

Would be simpler and we can just say "if the env var is set" rather than worrying about a particular value.

shotgun_api3/shotgun.py
```diff
@@ -1265,7 +1266,9 @@ def reset_user_agent(self):
         if self.client_caps.platform is not None:
             ua_platform = self.client_caps.platform.capitalize()
         self._user_agents = ["shotgun-json (%s)" % __version__,
-                             "Python %s (%s)" % (self.client_caps.py_version, ua_platform)]
+                             "Python %s (%s)" % (self.client_caps.py_version, ua_platform),
+                             "ssl-validation %s" % (int(self.config.no_ssl_validation is False))]
```
robblau: I thought we talked about including the openssl version if available?

Would a string like "ssl 0.9.8a (validate)" make sense?  And if we're on an older Python that doesn't have access "ssl unknown (novalidate)" or something like that?

shotgun_api3/shotgun.py
```diff
@@ -2164,6 +2176,27 @@ def _make_call(self, verb, path, body, headers):
             attempt += 1
             try:
                 return self._http_request(verb, path, body, req_headers)
+            except SSLHandshakeError, e:
+                # Not a SHA-2 error or SHA-2 error we want to enforce. SHA-2 errors look like:
```
robblau: It would be good to say what is going on here or at least refer to the warning message below in the comment.  So if somebody is just reading the comments they see what the code will do.

shotgun_api3/shotgun.py
```diff
@@ -2164,6 +2176,27 @@ def _make_call(self, verb, path, body, headers):
...
                 return self._http_request(verb, path, body, req_headers)
+            except SSLHandshakeError, e:
+                # Not a SHA-2 error or SHA-2 error we want to enforce. SHA-2 errors look like: 
+                #
+                #   [Errno 1] _ssl.c:480: error:0D0C50A1:asn1 encoding routines:ASN1_item_verify:
+                #   unknown message digest algorithm
+                #
+                # valid values should be "0" and "1" but we're trying to catch edge cases too
+                if not str(e).endswith("unknown message digest algorithm") or \
+                   os.environ.get("SHOTGUN_FORCE_CERTIFICATE_VALIDATION") not in [0, "0", "", None]:
```
robblau: Again for env var, just use
```python
if "SHOTGUN_FORCE_CERTIFICATE_VALIDATION" in os.environ:
```
?

shotgun_api3/shotgun.py
```diff
@@ -2164,6 +2176,27 @@ def _make_call(self, verb, path, body, headers):
...
             try:
                 return self._http_request(verb, path, body, req_headers)
+            except SSLHandshakeError, e:
+                # Not a SHA-2 error or SHA-2 error we want to enforce. SHA-2 errors look like: 
+                #
+                #   [Errno 1] _ssl.c:480: error:0D0C50A1:asn1 encoding routines:ASN1_item_verify:
+                #   unknown message digest algorithm
+                #
+                # valid values should be "0" and "1" but we're trying to catch edge cases too
+                if not str(e).endswith("unknown message digest algorithm") or \
```
robblau: Should test if we've already turned off validation so we don't end up warning three times if things keep failing with certs disabled.

shotgun_api3/shotgun.py
kporangehat: I agree. Was responding to a pass comment from Manne about this being a bit too uncertain as so what the acceptable value(s) are as well as differences bewteen how platforms handle them. But in this case, the value doesn't matter... so I will happily change this :)

shotgun_api3/shotgun.py
```diff
@@ -66,19 +66,20 @@
...
 NO_SSL_VALIDATION = False
 try:
-    import ssl
-    if os.environ.get("SHOTGUN_DISABLE_SSL_VALIDATION", False):
+    import ssl        
+except ImportError, e:
+    if "SHOTGUN_FORCE_CERTIFICATE_VALIDATION" in os.environ:
+        raise ImportError("%s. SHOTGUN_FORCE_CERTIFICATE_VALIDATION environment variable prevents "
+                          "disabling SSL certificate validation." % e)
+    else:
```
robblau: No need for the else since the if body ends with a raise.

kporangehat: if `ssl` fails to import and you don't have `SHOTGUN_FORCE_CERTIFICATE_VALIDATION` in your environment variables, the else will kick in and we want to disable cert validation right?

robblau: Yes, but that would happen if we just take that block of code and put it up one indentation level.  Just the else line, not the logic in the else block.

shotgun_api3/shotgun.py
```diff
@@ -217,10 +218,27 @@ def __init__(self):
...
+        ssl_version_str = None
+        ssl_version_num = "unknown"
+        try:
+            ssl_version_str = ssl.OPENSSL_VERSION
+        except AttributeError, NameError:
+            pass
+        if ssl_version_str:
+            try:
+                ssl_version_num = re.search("openssl\s*([\S.]+).*", ssl_version_str, 
+                                            flags=re.IGNORECASE).group(1)
```
robblau: Any worries about relying on the formatting of the SSL version string here?  Could we use it without parsing so that if there are changes in the future it doesn't break things?

shotgun_api3/shotgun.py
```diff
@@ -2164,6 +2200,37 @@ def _make_call(self, verb, path, body, headers):
...
+                
+                if self.config.no_ssl_validation is False:
+                    LOG.warning("SSLHandshakeError: this Python installation is incompatible with "
+                                "certificates signed with SHA-2. Disabling certificate validation. "
+                                "For more information, see http://blog.shotgunsoftware.com/2016/01/"
+                                "important-ssl-certificate-renewal-and.html")
+                self._close_connection()
+                self._turn_off_ssl_validation()
+                # reload user agent to reflect that we have turned off ssl validation
+                req_headers["user-agent"] = "; ".join(self._user_agents)
```
robblau: Should the few lines above (maybe not the close connection) be in the body of that if?  Should we do that each time through the loop or just the first?

shotgun_api3/shotgun.py
kporangehat: ah duh gotcha.

shotgun_api3/shotgun.py
kporangehat: the first will suffice. good one.

kporangehat:
merged and closed by 3e72b44

PR #98, opened 2016-01-18
For #34126 -Added check for image field in batch calls, then resolves…

Zaltu:
… them seperately


kporangehat:
Hey @Zaltu,

Looks like your merge clobbered a bunch of the latest SSL commits from master which is now causing the tests to fail. Might want to take a closer look and ensure those changes aren't getting reverted. 

cheers,
kp

shotgun_api3/shotgun.py
```diff
@@ -804,16 +804,16 @@ def create(self, entity_type, data, return_fields=None):
...
+        upload_data.append((data.pop('sg_uploaded_movie', None), 'sg_uploaded_movie'))
+        upload_data.append((data.pop('image', None), 'image'))
 
-        upload_filmstrip_image = None
         if 'filmstrip_image' in data:
             if not self.server_caps.version or self.server_caps.version < (3, 1, 0):
                 raise ShotgunError("Filmstrip thumbnail support requires server version 3.1 or "\
                     "higher, server is %s" % (self.server_caps.version,))
-            upload_filmstrip_image = data.pop('filmstrip_image')
+            upload_data.append((_safe_pop(data, 'filmstrip_image'), 'filmstrip_image'))
```
pboucher: Remove _safe_pop from here (use pop's default argument).

Is _safe_pop still defined somewhere? If not, this line should fail in tests. If this line doesn't fail, that means this code path isn't accounted for in the tests and we should edit / add tests to cover this. Which would mean having a Shotgun server with a version below 3.1.0 or using the Shotgun Mock to spoof a Shotgun server with a version below 3.1.0

shotgun_api3/shotgun.py
```diff
@@ -854,15 +852,17 @@ def update(self, entity_type, entity_id, data):
...
+
+        upload_data.append((data.pop('sg_uploaded_movie', None), 'sg_uploaded_movie'))
+        upload_data.append((data.pop('image', None), 'image'))
+
         if 'filmstrip_image' in data:
             if not self.server_caps.version or self.server_caps.version < (3, 1, 0):
                 raise ShotgunError("Filmstrip thumbnail support requires server version 3.1 or "\
                     "higher, server is %s" % (self.server_caps.version,))
-            upload_filmstrip_image = data.pop('filmstrip_image')
+            upload_data.append((_safe_pop(data, 'filmstrip_image'), 'filmstrip_image'))
```
pboucher: Remove _safe_pop and use pop with default arg.

Zaltu: Looks like I never pushed my last commit. Sorry.

PR #101, opened 2016-01-27
For #34500 : Cleaning up tests so travis-ci can automate them all

neilvictorgrey:
* Remove deprecated --use-mirrors flag from pip install
* Update .travis.yml to run all test files in the project
* Using .lower() from self.base_url breaks base64 encoding. Essentially, if users pass their credentials as part of the URL, those credentials will get set lower-case before being passed to the "Authorization" header. Handling this by just using base_url as it's passed in when checking for authorization parameters prefixed onto the URL.
* Allowing for environment variables to be used in addition to a configuration file. If an environment variable exists it will take precedence over the configuration file.  Environment variables are named the same as their configuration file counterparts, but they are uppercased and prefixed with SG_
* Added pertinent environment variables to our travis-ci setup so that we're good to go once this is merged. API keys and passwords are securely stored on their site.


shotgun_api3/shotgun.py
```diff
@@ -409,7 +409,7 @@ def __init__(self,
         else:
             self.__ca_certs = os.environ.get('SHOTGUN_API_CACERTS')
 
-        self.base_url = (base_url or "").lower()
+        self.base_url = (base_url or "")
```
manneohrstrom: I am worried that this may introduce regressions. Any client code which compares `sg.base_url` with something will suddenly compare against a potential new value. What's the benefit of removing `lower()`? Urls are case insensitive according to https://tools.ietf.org/html/rfc3986 (page 16):

```
Scheme names consist of a sequence of characters beginning with a
   letter and followed by any combination of letters, digits, plus
   ("+"), period ("."), or hyphen ("-").  Although schemes are case-
   insensitive, the canonical form is lowercase and documents that
   specify schemes must do so with lowercase letters.  An implementation
   should accept uppercase letters as equivalent to lowercase in scheme
   names (e.g., allow "HTTP" as well as "http") for the sake of
   robustness but should only produce lowercase scheme names for
   consistency.
```

tests/test_client.py
```diff
@@ -240,7 +240,7 @@ def test_rpc_error(self):
 
         try:
             self.sg.info()
-        except api.Fault, e:
+        except api.Fault as e:
```
manneohrstrom: will this break clients running the python API with py24/py25?

tests/base.py
```diff
@@ -300,13 +301,13 @@ def __init__(self):
...
         self.task_content   = None
 
-
     def read_config(self, config_path):
         config_parser = ConfigParser()
         config_parser.read(config_path)
         for section in config_parser.sections():
             for option in config_parser.options(section):
-                value = config_parser.get(section, option)
+                value = (os.environ.get('SG_{0}'.format(str(option).upper())) or
```
manneohrstrom: A comment here would help future code explorers. :) Adding a doc string to read_config I think would be useful too.

tests/test_client.py
neilvictorgrey: Ya, I didn't recall that py24/25 don't support "as"... it's been so long. This goes into way deeper than this PR, but can we not just say "we're deprecating support for py24/25 in future versions of the API"? I'm sure any studio using those versions are very much aware that they're dead and burried--- more-so, it puts an unnecessary support burden on us in needing to maintain a version of Python that's been deprecated since before Shotgun even was a product.

tests/base.py
neilvictorgrey: I'm breaking this up a bit more in the next commit, since getting the environment variables here is dependent on whether someone has setup a config file properly (or has one at all).

shotgun_api3/shotgun.py
neilvictorgrey: There's some reasoning further down in the code, but I agree that we shouldn't risk the larger potential issues by changing it here.  I'm reverting this and making a more specific change to where it's particularly relevant and almost zero risk.

manneohrstrom:
CR good to go! :+1:

tests/base.py
```diff
@@ -285,29 +286,34 @@ def setUp(self):
...
-        self.shot_code      = None
-        self.task_content   = None
 
+        for key in self.config_keys():
+            value = None
+            if key in ['mock']:
+                value = True
+            # Look for any environment variables that match our test
+            # configuration naming of "SG_{KEY}". Default is None.
+            value = os.environ.get('SG_{0}'.format(str(key).upper()))
```
robblau: Any worry about this having to run on older Pythons?

manneohrstrom: yes - well spotted! `format()` is 2.6+

neilvictorgrey: I'm too cool for school apparently-- will update that to % syntax.
As well @pboucher mentioned to do a check for whether "mock" is being passed in as an env var and change that to a boolean.
Will update those now.

tests/base.py
```diff
@@ -285,29 +286,34 @@ def setUp(self):
...
-        self.shot_code      = None
-        self.task_content   = None
 
+        for key in self.config_keys():
+
+            # Look for any environment variables that match our test
+            # configuration naming of "SG_{KEY}". Default is None.
+            value = os.environ.get('SG_%s' % (str(key).upper()))
+            if key in ['mock']:
+                value = (value == None) or (str(value).lower() in ['true','1']):
```
pboucher: Looks like a colon too many.

neilvictorgrey: ya, got too excited and should've run my tests locally first...

neilvictorgrey:
@robblau @manneohrstrom @pboucher any further feedback?

Also, I'm not familiar with the release process on this repo, so anyone who's got a few minutes to walk me through it -- your help would be much appreciated!

manneohrstrom:
@kporangehat usually does the final signoff and release prep when we do a release

PR #104, opened 2016-02-09
For #34814 - update test to reflect new server response

leweryo:


pboucher:
Good to merge when Travis tests return.

PR #105, opened 2016-02-10
For #33172, support ampersand in proxy authentication passwords

jfboismenu:
Splitting the string on @ was the right idea, but we have to do it from the end, since the password might have @ in it. Splitting from the end is safe since the the host and port number can't have this character.


manneohrstrom:
i like the brevity of this bug fix :)

kporangehat:
Yes short and sweet! My only comment is about the title and commit message. It's not an ampersand, it's an "at" symbol. An ampersand is `&`. When squashing, would be good to fix that. Otherwise merge away!

PR #108, opened 2016-02-26
For #35229: Added support for 'additional_server_presets' in read req…

daigles:
…uests


shotgun_api3/shotgun.py
```diff
@@ -186,6 +186,12 @@ def ensure_per_project_customization(self):
             'label': 'project parameter'
         }, True)
 
+    def ensure_support_for_additional_server_presets(self):
+        """Wrapper for ensure_support"""
+        return self._ensure_support({
+            'version': (6, 3, 11),
+            'label': 'project parameter'
```
robblau: Label needs to be updated.

shotgun_api3/shotgun.py
```diff
@@ -593,6 +599,9 @@ def find(self, entity_type, filters, fields=None, order=None,
         :param include_archived_projects: Optional, flag to include entities
         whose projects have been archived
 
+        :param additional_filter_presets: Optional, list of filter presets to
+        apply.
+
```
robblau: This should link to the docs for how to use filter presets.

shotgun_api3/shotgun.py
daigles: yes! Plan is to update before merging...

shotgun_api3/shotgun.py
manneohrstrom: I think small example would also be handy, like in doc string for the `order` parameter!

shotgun_api3/shotgun.py
```diff
@@ -675,6 +688,32 @@ def _construct_read_parameters(self,
...
+                if not isinstance(preset, list):
+                    msg = "Elements of 'additional_server_presets' parameter must be lists"
+                    raise ValueError(msg)
+
+                if len(preset) == 0:
+                    msg = "Missing 'name' for 'additional_server_presets'"
+                    raise ValueError(msg)
+
+                preset_param = {
+                    'name': preset[0]
```
manneohrstrom: please use double quotes

shotgun_api3/shotgun.py
```diff
@@ -675,6 +688,32 @@ def _construct_read_parameters(self,
         params["paging"] = { "entities_per_page": self.config.records_per_page,
                              "current_page": 1 }
 
+        if additional_server_presets is not None:
+            if not isinstance(additional_server_presets, list):
```
manneohrstrom: Nit: May want to allow tuples too, e.g.: `if not isinstance(additional_server_presets, (list, tuple)):`

shotgun_api3/shotgun.py
```diff
@@ -675,6 +688,32 @@ def _construct_read_parameters(self,
...
+                msg = "'additional_server_presets' parameter must be a list or None"
+                raise ValueError(msg)
+
+            preset_list = []
+            for preset in additional_server_presets:
+                if not isinstance(preset, list):
```
manneohrstrom: Also consider allowing tuples here.

manneohrstrom: For future code readers, adding a comment with an example of what the code is trying to parse would be helpful I think! E.g something along the lines of 
```
# now generate crud-ready filter preset
# ['name_of_preset', {'param1': 'value1', 'param2': 'value2'} ]
# ->
# {'name': 'name_of_preset', 'param1': 'value1', 'param2': 'value2'}
```

shotgun_api3/shotgun.py
manneohrstrom: Also, might be useful to add a small note here that presets get ANDed together with the rest of the filters?

shotgun_api3/shotgun.py
```diff
@@ -675,6 +688,32 @@ def _construct_read_parameters(self,
...
+                    msg = "Elements of 'additional_server_presets' parameter must be lists"
+                    raise ValueError(msg)
+
+                if len(preset) == 0:
+                    msg = "Missing 'name' for 'additional_server_presets'"
+                    raise ValueError(msg)
+
+                preset_param = {
```
manneohrstrom: nit: Maybe this variable could be named `crud_preset` or something? It's a little confusing that it's named preset_param but then internally contains a 'param's dictionary with the preset params :)

shotgun_api3/shotgun.py
```diff
@@ -675,6 +688,32 @@ def _construct_read_parameters(self,
...
+                    'name': preset[0]
+                }
+
+                if len(preset) > 1:
+                    preset_param['params'] = preset[1]
```
manneohrstrom: add an else clause with an empty dictionary for consistency?

shotgun_api3/shotgun.py
manneohrstrom: Actually, might be even better to break out this resolve logic into a separate method? Then the above suggested comment could just become the docstring.

shotgun_api3/shotgun.py
manneohrstrom: Qucick check at this point that `preset[1]` is a dict?

manneohrstrom:
CR done! Minor comments only from me!

shotgun_api3/shotgun.py
robblau: The string with the label says ```project parameter```.  It should be something for the presets.  Just making sure that's the bit you were planning on updating before a merge.

shotgun_api3/shotgun.py
robblau: Is there a reason to do this kind of a check at all?

Can do a try around the for loop or something else that would be a better way to support anything that is itterable, or just do nothing (much like the way sort_order is handled).  This kind of argument checking is a little un-pythonic where as long as a class implements the right combo of magic methods you let it pretend to be a list, a dict, or whatever.

shotgun_api3/shotgun.py
robblau: Agree with having an example in the comments.  We'll obviously need to add this syntax into the docs on what the filters look like in the wiki.

As above, I'd just leave out the isinstance checking.  No reason to stop people from using their own classes to wrap this kind of syntax.

shotgun_api3/shotgun.py
robblau: Again, I'd leave out the explicit check with isinstance.

shotgun_api3/shotgun.py
robblau: Should there be a warning or other kind of indication if len > 2?

shotgun_api3/shotgun.py
daigles: My goal was to validate user input as much as possible to guide users through parameter miss use.
But I'm all for keeping everything consistent so I'll follow your lead and remove the check altogether.

shotgun_api3/shotgun.py
daigles: Warning? as in "raise ValueError" or are you thinking of something else?

robblau: I was thinking of something like this:
https://github.com/shotgunsoftware/python-api/blob/35229_add_additional_server_presets/shotgun_api3/shotgun.py#L2255

daigles: ah! Thanks! I'll do that!

shotgun_api3/shotgun.py
```diff
@@ -186,6 +186,12 @@ def ensure_per_project_customization(self):
             'label': 'project parameter'
         }, True)
 
+    def ensure_support_for_additional_server_presets(self):
```
robblau: Should be "additional_filter_presets".

shotgun_api3/shotgun.py
```diff
@@ -615,13 +632,16 @@ def find(self, entity_type, filters, fields=None, order=None,
             # So we only need to check the server version if it is False
             self.server_caps.ensure_include_archived_projects()
 
+        if additional_filter_presets:
+            self.server_caps.ensure_support_for_additional_server_presets()
```
robblau: Should be "additional_filter_presets".

shotgun_api3/shotgun.py
```diff
@@ -693,6 +716,37 @@ def _construct_read_parameters(self,
...
+
+        :param additional_filter_presets: list of filter presets to process
+        :param params: param object to update
+        :return: The updated params object
+        """
+        if additional_filter_presets:
+            preset_list = []
+            for preset in additional_filter_presets:
+                if len(preset) == 0:
+                    msg = "Missing 'name' for 'additional_server_presets'"
```
robblau: Should be "additional_filter_presets".

tests/test_api.py
```diff
@@ -2050,198 +2050,296 @@ def test_simple(self):
...
         self.assertEqual(matches[0]["name"], "%s Text Search 3" % self._prefix)
         self.assertEqual(matches[1]["type"], "Asset")
         self.assertEqual(matches[1]["name"], "%s Text Search 4" % self._prefix)
-         
-        
 
 
+class TestReadAdditionalFilterPresets(base.LiveTestBase):
+    """
+    Unit tests for the additional_server_presets read parameter
```
robblau: Should be "additional_filter_presets".

shotgun_api3/shotgun.py
```diff
@@ -693,6 +716,37 @@ def _construct_read_parameters(self,
             params['sorts'] = sort_list
         return params
 
+    def _add_filter_presets(self, additional_filter_presets, params):
```
mattdaw: I found it confusing to have so many different uses of the name "params" in this method. What do you think about not passing in params to _add_filter_presets? Then it could instead return a dict that is merged back into params at line 699?

shotgun_api3/shotgun.py
```diff
@@ -564,7 +570,7 @@ def find_one(self, entity_type, filters, fields=None, order=None,
 
     def find(self, entity_type, filters, fields=None, order=None,
             filter_operator=None, limit=0, retired_only=False, page=0,
-            include_archived_projects=True):
+            include_archived_projects=True, additional_filter_presets=None):
```
robblau: Should this be added to find_one as well?

shotgun_api3/shotgun.py
```diff
@@ -593,6 +599,17 @@ def find(self, entity_type, filters, fields=None, order=None,
         :param include_archived_projects: Optional, flag to include entities
         whose projects have been archived
 
+        :param additional_filter_presets: Optional list of presets to
+        further filter the result set, list has the form:
+        [ { preset_name: <preset_name>, <optional_preset parameters> } ],
```
robblau: For python it should be something more like this for clarity and the fact that keys need quotes:
[{"preset_name": <preset_name>, <optional_param1>: <optional_value1>, ... }]

shotgun_api3/shotgun.py
```diff
@@ -593,6 +599,17 @@ def find(self, entity_type, filters, fields=None, order=None,
...
+
+        Note that these filters are ANDed together and ANDed with the 'filter'
+        argument.
+
+        For details on supported presets and the format of this parameter,
+        please consult the API documentation:
+        https://github.com/shotgunsoftware/python-api/wiki
```
robblau: When we do document this I believe it'll be at the following link, so we should probably be specific and not make somebody reading this hunt around:
https://github.com/shotgunsoftware/python-api/wiki/Reference%3A-Filter-Syntax

tests/test_api.py
```diff
@@ -2050,198 +2050,296 @@ def test_simple(self):
...
+class TestReadAdditionalFilterPresets(base.LiveTestBase):
+    """
+    Unit tests for the additional_filter_presets read parameter
+    """
+
+    def test_simple_case(self):
+
+        filters = [
+            ['project', 'is', self.project],
+            ['id', 'is', self.version['id']]
```
robblau: Should switch to double quotes throughout the tests (to be in line with our python coding conventions).

daigles: Yeah I was wondering about that. I went ahead and changed the entire test file. Can I do all of them?

shotgun_api3/shotgun.py
daigles: good point! Added!

shotgun_api3/shotgun.py
daigles: actually, I'm adding a new section for it so (I think) the link is going to be:
https://github.com/shotgunsoftware/python-api/wiki/Reference%3A-Additional_Filter_Presets-Syntax
Made the change. Will keep an eye on it and will test/update once the wiki changes are committed.

tests/test_api.py
robblau: I'm for changing all of them, but maybe outside of this pull request so that the diff doesn't end up all messy?

daigles: ok, fare enough. I'll fix the one in the new tests and open a ticket for the quotes... that I'll fix asap.

PR #111, opened 2016-04-05
For #10692: Option to use add/remove/set modes when updating multi-entity fields

neilvictorgrey:
This update will allow for updating multi-entity fields without having to pull down the field's entire dataset first.

It uses the new parameter ```multi_entity_update_modes``` on the update method to specify one of 'add', 'remove', or 'set'.

As this update is tied to a change on the server-side, it should not be merged until that pull request is merged and released.

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -270,7 +270,6 @@ class _Config(object):
         self.session_token = None
         self.authorization = None
         self.no_ssl_validation = False
-        self.multi_entity_update_modes = ['set','add','remove']
 
 
 class Shotgun(object):
@@ -851,9 +850,15 @@ class Shotgun(object):
 
         :param data: Required, dict fields to update on the entity.
 
-        :param multi_entity_update_modes: Optional, dict of what update mode to use when updating
-        a multi-entity link field.  The keys in the dict are the fields to set the mode for and
-        the values from the dict are one of "set", "add", or "remove".
+        :param multi_entity_update_modes: Optional, dict of what update mode to
+        use when updating a multi-entity link field.  The keys in the dict are
+        the fields to set the mode for and the values from the dict are one
+        of "set", "add", or "remove". The default behaviour is to use "set" if a
+        field being updated doesn't have an explicitly set mode. For example,
+        to append to the 'shots' field and remove from the 'assets' field on
+        a 'Sequence', you would specify:
+
+            multi_entity_update_modes={"shots":"add", "assets":"remove"}
 
         :returns: dict of the fields updated, with the entity_type and
         id added.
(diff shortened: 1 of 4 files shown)


shotgun_api3/shotgun.py
```diff
@@ -270,6 +270,7 @@ def __init__(self):
         self.session_token = None
         self.authorization = None
         self.no_ssl_validation = False
+        self.multi_entity_update_modes = ['set','add','remove']
```
robblau: I'd leave out the client side validation.  These will be checked server side and if the set of update modes ever changes on the server there will be an issue sync'ing up the python code.

shotgun_api3/shotgun.py
```diff
@@ -850,6 +851,10 @@ def update(self, entity_type, entity_id, data):
 
         :param data: Required, dict fields to update on the entity.
 
+        :param multi_entity_update_modes: Optional, dict of what update mode to use when updating
+        a multi-entity link field.  The keys in the dict are the fields to set the mode for and
+        the values from the dict are one of "set", "add", or "remove".
```
robblau: We should include an example value so the format of the argument is obvious.

neilvictorgrey: Yup, definitely agree. Will add that in.

shotgun_api3/shotgun.py
neilvictorgrey: My thinking here was that by specifying client-side, we could:

- Be more explicit as to which values are accepted.
- Do validation checks before reaching the server, effectively reducing time to failure and adding redundancy.
- Provide an error message that's very specific to these modes. (Although arguably we should just do better error messaging for this case server-side anyway!)
- The likelihood of modes being updated frequency is low.

While it does seem there are some benefits client-side, I do agree this code should be as generic as possible. Will update.

neilvictorgrey:
@robblau removed assumed validation and added example to docstring. Let me know if you see anything else that should be updated.

shotgun_api3/shotgun.py
```diff
@@ -850,6 +850,16 @@ def update(self, entity_type, entity_id, data):
 
         :param data: Required, dict fields to update on the entity.
 
+        :param multi_entity_update_modes: Optional, dict of what update mode to
+        use when updating a multi-entity link field.  The keys in the dict are
+        the fields to set the mode for and the values from the dict are one
+        of "set", "add", or "remove". The default behaviour is to use "set" if a
+        field being updated doesn't have an explicitly set mode. For example,
```
robblau: This is not very clear.  Maybe ```The default behavior if mode is not set for a field is "set".```?

robblau:
@neilvictorgrey Looking good.  Just one comment about a docstring.

PR #114, opened 2016-04-25
For #36143: Explicitly closing download_attachment file handler

neilvictorgrey:
We want to ensure that we're explicitly closing our file handler in download_attachment, so that we don't throw an error in cases where Python's garbage collector doesn't close the file.


neilvictorgrey:
@robblau @pboucher mind taking a look at this update?

robblau:
Had one comment for Neil out of band.  About the top close looking redundant.  He is following up.

PR #116, opened 2016-05-12
For #34163, better import strategy for mimetypes

jfboismenu:
We now always import the bundled `mimetypes`module on versions of Python that are broken.


pboucher:
LGTM!

PR #117, opened 2016-05-29
Convert doc-strings for sphinx and add standalone sphinx docs.

kporangehat:
This is a large commit almost entirely focused on the doc-strings and related formatting in order to release sphinx docs to replace the wiki.

This also adds some additional consistency to the doc-sting format and layout as well to conform to standards and internal style-guide.

A live preview is available at http://developer.shotgunsoftware.com/python-api/


docs/usage_tips.rst
```diff
@@ -0,0 +1,209 @@
...
+=========
+
+- ``name``: This is a UI-only field that is a combination of the ``firstname`` + ``' '`` + 
+  ``lastname``.
+
+Shot
+====
+
+**Smart Cut Fields**: These fields are available only in the browser UI. You can read more about 
+smart cut fields and the API in the Smart Cut Fields doc (link TK)::
```
robblau: Should actually link to the smart cut fields doc.

docs/tutorials.rst
```diff
@@ -0,0 +1,42 @@
...
+    examples/basic_find_shot
+    examples/basic_update_shot
+    examples/basic_delete_shot
+    examples/basic_create_shot_task_template
+    examples/basic_create_version_link_shot
+    examples/basic_upload_thumbnail_version
+
+********
+Advanced
+********
```
robblau: Are the advanced examples still being ported over?

README.md
```diff
@@ -33,13 +33,13 @@ If you're using pip with `requirements.txt`, add the following line:
...
 ## Documentation
-Tutorials and detailed documentation about the Python API are available on the [Python API Wiki](https://github.com/shotgunsoftware/python-api/wiki). There is also some additional related documentation on the [Shotgun Support Website](https://support.shotgunsoftware.com/forums/48807-developer-api-info).
+Tutorials and detailed documentation about the Python API are available on at http://developer.shotgunsoftware.com/python-api). There is also some additional related documentation on the [Shotgun Support Website](https://support.shotgunsoftware.com/forums/48807-developer-api-info).
 Some useful direct links in each are below:
 
 * [Introduction and Video Tour](https://www.youtube.com/watch?v=QdL5E9XbdJQ)
-* [Reference: Methods](https://github.com/shotgunsoftware/python-api/wiki/Reference%3A-Methods)
-* [Reference: Data Types](https://github.com/shotgunsoftware/python-api/wiki/Reference%3A-Data-Types)
-* [Reference: Filter Syntax](https://github.com/shotgunsoftware/python-api/wiki/Reference%3A-Filter-Syntax)
+* [API Reference](http://developer.shotgunsoftware.com/python-api/reference.html)
```
robblau: Possible to make the depth for reference one deeper so we can see methods in the TOC?  Might be too much once we see it, but it would be nice to have a simple list of what methods are there in one concise spot.

robblau:
Possible to update the TOC?  Currently getting this:

```
Installation
Authentication
API Usage Tips
Examples
API Reference
Filter Syntax
Data Types
Working With Files (Attachments)
Working With Local Files
Packaging an application with py2app (or py2exe)
Shotgun Server API Changelog
```

Seems like something along these lines would be a more intuitive ordering:

```
Installation
Authentication
API Reference
Filter Syntax
Data Types
Examples
API Usage Tips
Working with uploaded files
Working with local files
Packaging an application with py2app or py2exe
Shotgun Server API Changelog (deprecated)
```

README.md
kporangehat: After poking around a bit, I don't think this is doable without some significant hackery. I agree with you that it would be nice to have an index, but my vote is to punt on this for a bit.

kporangehat:
Ready for CR again. 

All docs from the wiki have been added. This version has been pushed to  http://developer.shotgunsoftware.com/python-api/

We can probably still improve the organization and update various things (like the svn integration docs... add a github version). But again, the goal here to to get to a point where we can shut down the wiki, then we can iterate on these to get them perfect. 🎉

manneohrstrom:
### Suggested feedback before roll-out

This looks amazing. My only real piece high level piece of feedback would be the TOC, which i don't think makes sense right now. It's too flat and it's hard to find things. How about this:

```
Installation
Authentication
API Reference
-- filter syntax
-- Data types
-- Event types
Cookbook
-- API Usage tips
-- Examples
-- Working with files
-- Workfile with local files
-- working with tasks
-- smart cut fields
Advanced
-- Packaging with py2exe
-- Ironpython
-- Server API changelog
```

The only other piece of feedback i have that I think we should address before an official transition is that the docs currently contain documentation for the internal classes `ServerCapabilties` and `ClientCapabilitites`. If they are part of the the official API that we promise never to change, then we should document them for sure. But as far as I understand, these are internal and not intended to be used directly. So they should not be in the sphinx docs. Instead, we should add a warning in the docstrings emphasizing that these classes as part of the internal API and their interfaces may change in the future. Documenting internal classes in the sphinx docs will be confusing to users.

### Suggested post-rollout tweaks

Other than that I think it looks amazing: I did a really quick readthrough and jotted down the following feedback - maybe some of this should be considered for a quick subsequent post-doc-transition pass? 

- Event types section should contain a link to the event daemon docs. (and vice versa!)

- Does smart cut fields deserve its own heading at all? Is it legacy at this point? I think we should at the very least add some notes on the smart cut fields page about sg7 to mention the cut schema.

- minor: in the shotgun server API section, it says "This list is out of date and has not been updated since 5/2014" - this sort of date format is confusing for non-americans. For maximum clarity, simply write "May 2014".

- What's the difference between 'working with files' and 'working with local files'? can the title be reworked for clarity?

- Installation: I think ironpython should be a topic in the advanced section.

- installation: I think most people install by downloading from github: So the documentation for how to do that should probably be the first item, before pip, so people don't have to scroll.

- Would be great with some more complex filter examples. Maybe these could be moved across into the "cookbook"

- Link from the filter presets param in the API reference docs should like to the filter presets doc section.

- I see things like "New in version 3.0.3: Requires Shotgun Server v2.2.0+" - do we really need this sort of info? Is ANYONE running SG v2.2.0 or less? Maybe just say that these docs cover SG v6 and above and do a pass to simplify?

- Do we need a ticket for a rundeck task (or something) that re-builds the sphinx docs for the python API?

- for the shotgun API reference, it would be great if we could divide the methods into subsections, like 'basic crud', working on the schema, domain-specific methods etc. Check out what josh did in sg utils: https://github.com/shotgunsoftware/tk-framework-shotgunutils/blame/3ad7c188ae582f52e105319a73ab76132ea1dc8e/docs/shotgun_hierarchy_model.rst#L158

- This wording struck me as a little odd: _Shotgun generates read-only EventLogEntries for every destructive action performed by a user. And by “destructive”, we simply mean an action that updates the database somehow, not some tantrum that a user has (hopefully)._ 
Would it be clearer to say something like: _Whenever a user makes a change to any data in Shotgun, an event log entry record is created, capturing the value before and after._?

kporangehat:
@manneohrstrom great notes! What were you referring to by this one?

> Would be great with some more complex filter examples. Maybe these could be moved across into the "cookbook"

manneohrstrom:
@kporangehat in the filter syntax section, you currently have this:

image

I think what would be great would be to:

- move the examples into the "cookbook" section and put a reference in the docs so that people can jump easily from the reference docs (e.g. the location above) to the cookbook.

- in the cookbook, have a collection of examples, showing different ways to combine things

- maybe add a number of example both of basic and advanced syntax.

- what would be ideal is if there is a cut'n'paste ready example for most common patterns, so that a TD could quickly grab some code straight out of the docs and transform it into something specifically for them.

- examples with nested query syntax would be great with the cookbook syntax. Also a screenshot showing how you can hover over a location in the shotgun fields list and see its nested API syntax. This is a lesser known feature and it's extremely useful. 

- In fact, in general it might be great to have a section to answer the following
  - Where do i go to see what entity types I can use in the API?
  - How do i find the name of a field i want to use?

kporangehat:
Pushed and built a new version at http://developer.shotgunsoftware.com/python-api/

Addressed most of the things you brought up above (marked with ✅ ). 

**Outstanding items:**

- [ ] **Complex filter examples as well as additional "cut-n-paste" examples in the cookbook.**
       
       _Neil is working on some of these and I'll coordinate with him to get them added to the right place. I think that the single examples were relevant still in the doc so I left them there for now._
    
- [ ] **New in version / Requires Shotgun Server info updates.**

    _I added these as I went and some of them got very irrelevant but I kept them in for the time being. We can remove stuff that's old enough. I like the idea here, but making it actually relevant would be better._

- [ ] **Automated builds of the docs.**

    _Working on this one._

- [ ] **FAQ section for things like how to get the system name / dot-notation of linked fields, etc.**

    _Perhaps we can start compiling a list._

- [ ] **Lots of additional cleanup and tweaks** 🚜

kporangehat:
What do you think about moving that changelog to its own CHANGELOG file?

-- 
KP
Shotgun | Autodesk

From: Rob Blau <notifications@github.com> <notifications@github.com>
Reply: shotgunsoftware/python-api
<reply@reply.github.com>
<reply@reply.github.com>
Date: September 19, 2016 at 7:38:51 PM
To: shotgunsoftware/python-api <python-api@noreply.github.com>
<python-api@noreply.github.com>
CC: KP <kp@shotgunsoftware.com> <kp@shotgunsoftware.com>, Mention
<mention@noreply.github.com> <mention@noreply.github.com>
Subject:  Re: [shotgunsoftware/python-api] Convert doc-strings for sphinx
and add standalone sphinx docs. (#117)

*@robblau* commented on this pull request.

robblau:
I'd be game for that. Feels pretty equivalent, but he docs should have an easy to find pointer to them.

PR #118, opened 2016-07-05
For #37361: Updating config to randomize human user

neilvictorgrey:
Randomizing the human user account created when running tests, unless otherwise specified by an environment variable.

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -211,8 +211,6 @@ class LiveTestBase(TestBase):
                 'password_proxy':self.config.human_password}
         if self.sg_version >= (3, 0, 0):
             data['locked_until'] = None
-
-
         self.human_user = _find_or_create_entity(self.sg, 'HumanUser', data)
 
         data = {'code':self.config.asset_code,
@@ -273,6 +271,10 @@ class LiveTestBase(TestBase):
                 'linux_path':'nowhere'}
         self.local_storage = _find_or_create_entity(self.sg, 'LocalStorage', data, keys)
 
+    def tearDown(self):
+        self.sg.delete("HumanUser", self.human_user["id"])
+        super(LiveTestBase, self).tearDown()
+
 
 class HumanUserAuthLiveTestBase(LiveTestBase):
     '''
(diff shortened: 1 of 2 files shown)


RobertLBelanger:
LGTM

neilvictorgrey:
Causing an issue when running multiple test sessions concurrently. Also, tearDown is creating a new user for every test, not every test session.

tests/base.py
```diff
@@ -273,6 +271,10 @@ def _setup_db(self, config):
                 'linux_path':'nowhere'}
         self.local_storage = _find_or_create_entity(self.sg, 'LocalStorage', data, keys)
 
+    def tearDown(self):
+        self.sg.delete("HumanUser", self.human_user["id"])
```
manneohrstrom: I suggest adding a comment here explaining why we are deleting the HumanUser but keeping all the other entities that we are creating as part of the setup. Also please add a docstring to all new methods that are introduced into the codebase.

tests/base.py
```diff
@@ -296,14 +298,16 @@ class SgTestConfig(object):
...
-
             # Look for any environment variables that match our test
             # configuration naming of "SG_{KEY}". Default is None.
             value = os.environ.get('SG_%s' % (str(key).upper()))
             if key in ['mock']:
                 value = (value == None) or (str(value).lower() in ['true','1'])
             setattr(self, key, value)
 
+        if not getattr(self, "human_password", None):
+            setattr(self, "human_password", os.urandom(16).encode('hex'))
```
manneohrstrom: please add a comment here to explain why we are doing this.

manneohrstrom: why doing the check for if `human_password` is already set? Is it necessary? Will that case work? I suggest either remote it or explain in a comment.

manneohrstrom:
I added some notes.

manneohrstrom:
One more thing: Just noticed the description of the PR sayijng "unless otherwise specified by an environment variable." - this doesn't seem to be outlined or documented anywhere? Is the idea that you can set it optionally in the config file? In that case I think you should document it within the config file itself - something like this:

```
human_name     : Sg unittest human
human_login    : sgunittesthuman

# Note: for security reasons, passwords generated during unit testing
# are randomized. If you want to use a specific password for your testing,
# you can uncomment the line below:
#human_password : hard_coded_password
```

PR #122, opened 2016-09-01
#36383 Adds support for new `nav_expand()` call for querying SG hierarchy

josh-t:
Creating PR for work done by @daigles on `nav_expand()` endpoint...

* Adds method to ensure filter presets are supported
* adds optional `additional_filter_presets` args where needed with links to docs
* adds unit test for `nav_expand` call.


josh-t:
Manually merged.

PR #125, opened 2016-10-27
for #39129: catch empty files before uploading() and raise exception.

kporangehat:
It was possible to attempt to upload a zero-byte file which can cause all sorts of issues but most notably it generates a ServerError which is misleading to the user. We now check the file exists and is not empty before calling `upload()`. If the file is empty, and exception is now raised stating the issue.

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -2137,7 +2137,7 @@ class Shotgun(object):
 
         :param str entity_type: Entity type to link the upload to.
         :param int entity_id: Id of the entity to link the upload to.
-        :param str path: Full path to file on disk to upload.
+        :param str path: Full path to an existing non-empty file on disk to upload.
         :param str field_name: The internal Shotgun field name on the entity to store the file in.
             This field must be a File/Link field type.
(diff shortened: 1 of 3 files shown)


shotgun_api3/shotgun.py
```diff
@@ -2137,17 +2137,21 @@ def upload(self, entity_type, entity_id, path, field_name=None, display_name=Non
...
         :returns: Id of the Attachment entity that was created for the image.
         :rtype: int
         """
+        # Basic validations of the file to upload.
         path = os.path.abspath(os.path.expanduser(path or ""))
         if not os.path.isfile(path):
             raise ShotgunError("Path must be a valid file, got '%s'" % path)
+        if os.path.getsize(path) == 0:
+            raise ShotgunError("The file \"%s\" is empty. Please provide a non-empty file "
+                               "to upload" % path)
```
robblau: I'd just change this to be consistent with the tone of the other error:

```
Path cannot be an empty file: '%s'
```

robblau:
One little recommendation, but looks good in general.  Good to go from me.

PR #126, opened 2016-10-31
mockgun: add support of status_list

foutoucour:
We faced the following warnings from mockgun when we tried to use the status_list on shots:

* `ShotgunError: The not_in operator is not supported on the status_list type` (fixed in `_compare`)
* `ShotgunError: Field Shot.sg_status_list: Handling for Shotgun type status_list is not implemented` (fixed in `_validate_entity_data`)

Let me know if it should be done another way.

Best,
Jordi

### DIFF
--- a/shotgun_api3/lib/mockgun.py
+++ b/shotgun_api3/lib/mockgun.py
@@ -526,6 +526,7 @@ class Shotgun(object):
                                    "serializable": dict,
                                    "date": datetime.date,
                                    "date_time": datetime.datetime,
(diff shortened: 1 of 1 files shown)


foutoucour:
Hey Here,

It almost a month we are waiting for feedback on this.

Would you have any ETA?

Jordi

JeanChristopheMorin:
I also added the support of `list`

foutoucour:
@jfboismenu Hey could we have updates here?

PR #130, opened 2017-02-05
Adds filter operator support for Mockgun.

jfboismenu:
- Adds supports for the "all" and "any" filter operator syntax inside queries, which can also be used recursively.
- Adds unit tests for the new functionality.


shotgun_api3/lib/mockgun.py
```diff
@@ -328,24 +330,7 @@ def find(self, entity_type, filters, fields=None, order=None, filter_operator=No
...
-                # ["field", "in", 2] --> ["field", "in", [2]]
-                new_filter = [ f[0], f[1], [ f[2] ] ]
-
-            else:
-                new_filter = f
-
-            resolved_filters_2.append(new_filter)
-
-        results = [row for row in self._db[entity_type].values() if self._row_matches_filters(entity_type, row, resolved_filters_2, filter_operator, retired_only)]
+        results = [row for row in self._db[entity_type].values() if self._row_matches_filters(entity_type, row, resolved_filters, filter_operator, retired_only)]
```
josh-t: I see this isn't something you actually changed, but wondering if we can clean it up a bit?
```
results = [
    row for row in self._db[entity_type].values() 
        if self._row_matches_filters(
            entity_type, 
            row, 
            resolved_filters, 
            filter_operator, 
            retired_only
        )
]
```
Maybe throw some comments in there too?

shotgun_api3/lib/mockgun.py
```diff
@@ -690,36 +675,91 @@ def _get_field_type(self, entity_type, field):
         except ValueError:
             return self._schema[entity_type][field]["data_type"]["value"]
 
-    def _row_matches_filter(self, entity_type, row, filter):
-        
-        
+    def _row_matches_filter(self, entity_type, row, filter, retired_only):
```
josh-t: I know you didn't add this, but I recommend not using `filter` as a variable name since it is a python [built-in](https://docs.python.org/2/library/functions.html#filter).

shotgun_api3/lib/mockgun.py
```diff
@@ -690,36 +675,91 @@ def _get_field_type(self, entity_type, field):
...
+        rearranged_filters = []
+
+        # now translate ["field", "in", 2,3,4] --> ["field", "in", [2, 3, 4]]
+        for f in filters:
+            if isinstance(f, list):
+                if len(f) > 3:
+                    # ["field", "in", 2,3,4] --> ["field", "in", [2, 3, 4]]
+                    new_filter = [f[0], f[1], f[2:]]
+
+                elif f[1] == "in" and not isinstance(f[2], list):
```
josh-t: Is this guaranteed to be 3 items?

tests/test_mockgun.py
```diff
@@ -0,0 +1,264 @@
+#! /opt/local/bin/python
+
+#  -----------------------------------------------------------------------------
+#  Copyright (c) 2009-2015, Shotgun Software Inc
```
josh-t: `2017`?

shotgun_api3/lib/mockgun.py
jfboismenu: This is code that I have moved around, I've actually only added the `elif isinstance(f, dict)` branch. Honestly, the fact that I've just written the first unit tests for mockgun makes me a bit hesitant to update too much.

PR #131, opened 2017-02-06
For #40631: Adding following method and updating unit tests

neilvictorgrey:
Changes:

- Adding method `Shotgun.following`, which returns a list of entity instances a HumanUser is following. It can optionally be passed a project description and/or an entity type name. eg.:
  - `sg.following( {"type":"HumanUser", "id":1234}, project={"type":"Project", "id":1234}, 
    entity_type="Task" )`
- Adding method `ServerCapabilities.ensure_user_following_support`, which ensures we're not calling this function if the server-side doesn't support it.
- As `Follow` isn't treated like a regular entity:
  - `TestFollow.setUp` was updated to clear the server state of pre-existing follows before each test. 
  - `TestFollow.test_follow` & `TestFollow.test_unfollow` were combined into one method `TestFollow.test_follow_unfollow` to match this test flow.
- Added method `TestFollow.test_following` for new functionality.

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -238,6 +238,15 @@ class ServerCapabilities(object):
             'label': 'additional_filter_presets parameter'
         }, True)
 
+    def ensure_user_following_support(self):
+        """
+        Ensures server has support for listing items a user is following, added in v7.0.12.
+        """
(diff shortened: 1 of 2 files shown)


francoisjacques:
LGTM !

PR #132, opened 2017-02-06
For #38721: Added support for direct upload to S3

daigles:
- Only supports uploads of files attached to Versions for now.

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -582,6 +582,7 @@ class Shotgun(object):
         >>> sg.server_info
         {'full_version': [6, 3, 15, 0],
          's3_uploads_enabled': True,
+         's3_direct_uploads_enabled': True,
          'version': [6, 3, 15]}
 
         :returns: dict of server information from :class:`ServerCapabilities` object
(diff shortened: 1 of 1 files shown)


shotgun_api3/shotgun.py
```diff
@@ -2155,6 +2156,132 @@ def upload(self, entity_type, entity_id, path, field_name=None, display_name=Non
         is_thumbnail = (field_name in ["thumb_image", "filmstrip_thumb_image", "image",
                                        "filmstrip_image"])
 
+        # Thumbnails can only be uploaded to SG for now...
+        if self.server_info['s3_direct_uploads_enabled'] \
```
manneohrstrom: double quotes here for consistency

shotgun_api3/shotgun.py
```diff
@@ -2155,6 +2156,132 @@ def upload(self, entity_type, entity_id, path, field_name=None, display_name=Non
         is_thumbnail = (field_name in ["thumb_image", "filmstrip_thumb_image", "image",
                                        "filmstrip_image"])
 
+        # Thumbnails can only be uploaded to SG for now...
```
manneohrstrom: might be clearer to say something like "Review quicktimes are handled as a special case and uploaded directly to amazon S3"

shotgun_api3/shotgun.py
```diff
@@ -2206,6 +2333,58 @@ def upload(self, entity_type, entity_id, path, field_name=None, display_name=Non
...
+
+        :param bool is_thumbnail: indicates if the attachment is a thumbnail.
+        :param str filename: name of the file that will be uploaded.
+
+        :returns: dictionary containing the ulpload_info and the url to upload to.
+        :rtype: dict
+        """
+
+        if is_thumbnail:
+            upload_type = 'Thumbnail'
```
manneohrstrom: Double quotes throughout for consistency would be great.

shotgun_api3/shotgun.py
```diff
@@ -2206,6 +2333,58 @@ def upload(self, entity_type, entity_id, path, field_name=None, display_name=Non
...
+        }
+
+        params.update(self._auth_params())
+
+        upload_url = "/upload/api_get_upload_link_info"
+        url = urlparse.urlunparse((self.config.scheme, self.config.server,
+                                   upload_url, None, None, None))
+
+        # Create opener with extended form post support
+        opener = self._build_opener(FormPostHandler)
```
manneohrstrom: It seems we have this logic in a couple of places:

- https://github.com/shotgunsoftware/python-api/blob/master/shotgun_api3/shotgun.py#L2022
- https://github.com/shotgunsoftware/python-api/blob/master/shotgun_api3/shotgun.py#L2188

at a glance, seems there could be value in addin a helper method that could be used in all three places. (sorry if i am overlooking something!)

shotgun_api3/shotgun.py
```diff
@@ -2206,6 +2333,58 @@ def upload(self, entity_type, entity_id, path, field_name=None, display_name=Non
...
+                raise ShotgunError("Unanticipated error occurred uploading "
+                                   "%s: %s" % (path, e))
+        else:
+            if not str(upload_info).startswith("1"):
+                raise ShotgunError("Could not get upload_url but " \
+                                   "not sure why.\nPath: %s\nUrl: %s\nError: %s" % (
+                                       path, url, str(upload_info)))
+
+        LOG.debug("Completed rpc call to %s" % (upload_url))
+        return {
```
manneohrstrom: A comment with an example of the expected return data from sg would be helpful here

shotgun_api3/shotgun.py
```diff
@@ -2206,6 +2333,58 @@ def upload(self, entity_type, entity_id, path, field_name=None, display_name=Non
...
         return attachment_id
 
+    def _get_attachment_upload_info(self, is_thumbnail, filename):
+        """
+        Internal function to get the information needed to upload a file to Cloud storage.
+
+        :param bool is_thumbnail: indicates if the attachment is a thumbnail.
+        :param str filename: name of the file that will be uploaded.
+
+        :returns: dictionary containing the ulpload_info and the url to upload to.
```
manneohrstrom: ulpload_info (typo)

manneohrstrom: Is upload info the data that you are passing back to the server in "pass 3" of the upload process? Might be worth adding a sentence to explain what it is here.

shotgun_api3/shotgun.py
```diff
@@ -2155,6 +2156,132 @@ def upload(self, entity_type, entity_id, path, field_name=None, display_name=Non
...
+        filename = os.path.basename(path)
+
+        # Step 1: get the upload url
+
+        upload_info = self._get_attachment_upload_info(is_thumbnail, filename)
+
+        # Step 2: upload the file
+
+        # Get some file info
+        fd = open(path, "rb")
```
manneohrstrom: In order to not leak file handles, this should probably be 

```
fd = open(path, "rb")
try:
    # do stuff
finally:
    fd.close()
```

(note that we cannot use the `with` statement if we want to be compatible with python2.4)

shotgun_api3/shotgun.py
```diff
@@ -2155,6 +2156,132 @@ def upload(self, entity_type, entity_id, path, field_name=None, display_name=Non
...
+        content_type = mimetypes.guess_type(filename)[0]
+        content_type = content_type or "application/octet-stream"
+        file_size = os.fstat(fd.fileno())[stat.ST_SIZE]
```
manneohrstrom: note: might be clearer to just use `os.path.getsize(path)` here. And no need to open the file open operation in that case.

shotgun_api3/shotgun.py
```diff
@@ -2155,6 +2156,132 @@ def upload(self, entity_type, entity_id, path, field_name=None, display_name=Non
...
+
+        # Perform the request
+        opener = urllib2.build_opener(urllib2.HTTPHandler)
+        try:
+            file = open(path, "rb")
```
manneohrstrom:
```
fd = open(path, "rb")
try:
    ...
except urllib2.HTTPError, e:
    ...
finally:
    fd.close()
```
(also note - nitpicky - that file is a reserved word in python so better to use some other variable name)

So if we use `os.path.getsize(path)` as suggested above, we don't need to open the file until we get to this point in the code.

shotgun_api3/shotgun.py
```diff
@@ -2155,6 +2156,132 @@ def upload(self, entity_type, entity_id, path, field_name=None, display_name=Non
...
+            request = urllib2.Request(upload_info['upload_url'], data=fd)
```
manneohrstrom: this is confusing - here we are using `fd` rather than `file` - looks like we are opening the file twice!

shotgun_api3/shotgun.py
```diff
@@ -2155,6 +2156,132 @@ def upload(self, entity_type, entity_id, path, field_name=None, display_name=Non
...
+            return self._upload_to_storage(entity_type, entity_id, path, field_name, display_name,
+                                           tag_list, is_thumbnail)
+        else:
+            return self._upload_to_sg(entity_type, entity_id, path, field_name, display_name,
+                                      tag_list, is_thumbnail)
+
+    def _upload_to_storage(self, entity_type, entity_id, path, field_name, display_name,
+                           tag_list, is_thumbnail):
+        """
+        Internal function to upload a file to S3 and link it to the specified entity.
```
manneohrstrom: nitpick: given that we are being given an url from the server that we are uploading to, maybe we should just omit any mention of `S3`? AFAIK, as far as the shotgun API is concerned, it could be any sort of service and there is nothing S3 specific in the code right now other than comments and log messages.

shotgun_api3/shotgun.py
```diff
@@ -2155,6 +2156,132 @@ def upload(self, entity_type, entity_id, path, field_name=None, display_name=Non
...
+            request.add_header("Content-Length", file_size)
+            request.get_method = lambda: "PUT"
+            result = opener.open(request)
+        except urllib2.HTTPError, e:
+            if e.code == 500:
+                raise ShotgunError("Server encountered an internal error.\n%s\n%s\n\n" % (url, e))
+            else:
+                raise ShotgunError("Unanticipated error occurred uploading %s: %s" % (path, e))
+
+        LOG.debug("Completed upload of media to S3 of file: %s", filename)
```
manneohrstrom: nitpick (see above): Consider dropping 'S3'

shotgun_api3/shotgun.py
```diff
@@ -2155,6 +2156,132 @@ def upload(self, entity_type, entity_id, path, field_name=None, display_name=Non
...
+            # we allow linking to nothing for generic reference use cases
+            if field_name is not None:
+                params["field_name"] = field_name
+            params["display_name"] = display_name
+            # None gets converted to a string and added as a tag...
+            if tag_list:
+                params["tag_list"] = tag_list
+
+        # Create opener with extended form post support
+        opener = self._build_opener(FormPostHandler)
```
manneohrstrom: again, this could perhaps take advantage of a shared method for running the form post style operations

shotgun_api3/shotgun.py
daigles: darn... though I had all of them :\

shotgun_api3/shotgun.py
daigles: yes. Agreed. this was also bugging me. will change

shotgun_api3/shotgun.py
daigles: done

shotgun_api3/shotgun.py
daigles: we actually need the fd a few lines below so left this as is.

shotgun_api3/shotgun.py
daigles: updated...

shotgun_api3/shotgun.py
daigles: yup. corrected

shotgun_api3/shotgun.py
daigles: added a _send_form(url, params) method.

PR #134, opened 2017-02-13
Mockgun percent

foutoucour:
We needed to mock percent fields in some tests.

I would like to do some unittests on implementations as such. Do you have any advises?


jfboismenu:
Actually we do now. We've started fixing some bugs in Mockgun recently and as such we've created unit tests. The first few tests are [here](https://github.com/shotgunsoftware/python-api/blob/master/tests/test_mockgun.py)

To run the tests for mockgun only, simply python

`PYTHONPATH=.. python test_mockgun.py`

PR #135, opened 2017-02-15
For #34923, desktop site bootstrap

jfboismenu:
Unit tests in Toolkit were super slow. I realized that for our tests we always instantiate Mockgun, which loads the schema from disk. Now we're keeping that schema cached. Unit tests for Toolkit now 5 times faster on my Macbook Pro and have gone from 160 seconds to 34.

Also, I decided to start splitting up Mockgun into multiple files at the same time. It is therefore now a module with the expected `__init__.py` that recreates the original API.

### DIFF
--- a/shotgun_api3/lib/mockgun/schema.py
+++ b/shotgun_api3/lib/mockgun/schema.py
@@ -1,0 +1,119 @@
+"""
+ -----------------------------------------------------------------------------
+ Copyright (c) 2009-2017, Shotgun Software Inc
+
+ Redistribution and use in source and binary forms, with or without
+ modification, are permitted provided that the following conditions are met:
+
+  - Redistributions of source code must retain the above copyright notice, this
+    list of conditions and the following disclaimer.
+
+  - Redistributions in binary form must reproduce the above copyright notice,
+    this list of conditions and the following disclaimer in the documentation
+    and/or other materials provided with the distribution.
+
+  - Neither the name of the Shotgun Software Inc nor the names of its
+    contributors may be used to endorse or promote products derived from this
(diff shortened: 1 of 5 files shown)


shotgun_api3/lib/mockgun/schema.py
```diff
@@ -0,0 +1,124 @@
...
+    :param sg_url: Shotgun site url
+    :param sg_script: Script name to connect with
+    :param sg_key: Script key to connect with
+    :param schema_file_path: Path where to write the main schema file to
+    :param schema_entity_file_path: Path where to write the entity schema file to
+    """
+
+    schema = shotgun.schema_read()
+    try:
+        fh = open(schema_file_path, "w")
```
josh-t: Can you use `with` here? If the `open` call raises for whatever reason, `fh` won't be bound and your `finally` clause will also error.

shotgun_api3/lib/mockgun/schema.py
```diff
@@ -0,0 +1,124 @@
...
+        pickle.dump(schema, fh, protocol=_HIGHEST_24_PICKLE_PROTOCOL)
+    finally:
+        fh.close()
+
+    schema_entity = shotgun.schema_entity_read()
+    try:
+        fh = open(schema_entity_file_path, "w")
```
josh-t: Same as above.

shotgun_api3/lib/mockgun/schema.py
jfboismenu: You are going to laugh but... the Shotgun API is supported on Python 2.4, which doesn't support the with statement...

PR #136, opened 2017-02-15
For #38721: Fixed upload regression for sites prior to V 7.0.10

daigles:
@jfboismenu as discussed...


daigles:
@jfboismenu I definitely want to keep my changes in the API as 'pythonesque' as possible! Updated the change as you proposed!

PR #137, opened 2017-03-02
For #40884: Extend support for direct uploads to S3

daigles:
- Added multi-part uploads to support uploads of files > 5 GB

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -2059,10 +2059,6 @@ class Shotgun(object):
             "filmstrip_thumbnail" : filmstrip_thumbnail,
         }
 
-        params.update(self._auth_params())
-
-        # Create opener with extended form post support
-        opener = self._build_opener(FormPostHandler)
         url = urlparse.urlunparse((self.config.scheme, self.config.server,
             "/upload/share_thumbnail", None, None, None))
 
@@ -2219,33 +2215,16 @@ class Shotgun(object):
 
         # Step 1: get the upload url
 
-        upload_info = self._get_attachment_upload_info(is_thumbnail, filename)
+        multipart_upload = (os.path.getsize(path)  > 20000000)
+
+        upload_info = self._get_attachment_upload_info(is_thumbnail, filename, multipart_upload)
 
         # Step 2: upload the file
 
-        fd = open(path, "rb")
-        try:
-            content_type = mimetypes.guess_type(filename)[0]
-            content_type = content_type or "application/octet-stream"
-            file_size = os.fstat(fd.fileno())[stat.ST_SIZE]
-
-            # Perform the request
-            opener = urllib2.build_opener(urllib2.HTTPHandler)
-
-            request = urllib2.Request(upload_info["upload_url"], data=fd)
-            request.add_header("Content-Type", content_type)
-            request.add_header("Content-Length", file_size)
-            request.get_method = lambda: "PUT"
-            result = opener.open(request)
-        except urllib2.HTTPError, e:
-            if e.code == 500:
-                raise ShotgunError("Server encountered an internal error.\n%s\n%s\n\n" % (url, e))
-            else:
-                raise ShotgunError("Unanticipated error occurred uploading %s: %s" % (path, e))
-        finally:
-            fd.close()
-
-        LOG.debug("File uploaded to Cloud storage: %s", filename)
+        if multipart_upload:
+            self._multipart_upload_file_to_storage(path, upload_info)
+        else:
+            self._upload_file_to_storage(path, upload_info["upload_url"])
 
         # Step 3: create the attachment
 
@@ -2342,12 +2321,13 @@ class Shotgun(object):
         attachment_id = int(str(result).split(":")[1].split("\n")[0])
         return attachment_id
 
-    def _get_attachment_upload_info(self, is_thumbnail, filename):
+    def _get_attachment_upload_info(self, is_thumbnail, filename, multipart_upload):
         """
         Internal function to get the information needed to upload a file to Cloud storage.
 
         :param bool is_thumbnail: indicates if the attachment is a thumbnail.
         :param str filename: name of the file that will be uploaded.
+        :param bool multipart_upload: Indicates if we want multi-part upload information back.
 
         :returns: dictionary containing the upload url and
             upload_info (passed back to the SG server once the upload is completed).
@@ -2364,7 +2344,8 @@ class Shotgun(object):
             "filename" : filename
         }
 
-        params.update(self._auth_params())
+        if multipart_upload:
+            params['multipart_upload'] = True
 
         upload_url = "/upload/api_get_upload_link_info"
         url = urlparse.urlunparse((self.config.scheme, self.config.server,
@@ -2374,11 +2355,17 @@ class Shotgun(object):
         if not str(upload_info).startswith("1"):
             raise ShotgunError("Could not get upload_url but " \
                                "not sure why.\nPath: %s\nUrl: %s\nError: %s" % (
-                                   path, url, str(upload_info)))
+                                   filename, url, str(upload_info)))
 
         LOG.debug("Completed rpc call to %s" % (upload_url))
+
+        upload_info_parts = str(upload_info).split("\n")
+
         return {
-            "upload_url" : str(upload_info).split("\n")[1],
+            "upload_url" : upload_info_parts[1],
+            "timestamp" : upload_info_parts[2],
+            "upload_type" : upload_info_parts[3],
+            "upload_id": upload_info_parts[4],
             "upload_info" : upload_info
         }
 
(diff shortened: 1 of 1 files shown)


shotgun_api3/shotgun.py
```diff
@@ -2059,10 +2059,6 @@ def share_thumbnail(self, entities, thumbnail_path=None, source_entity=None,
             "filmstrip_thumbnail" : filmstrip_thumbnail,
         }
 
-        params.update(self._auth_params())
-
-        # Create opener with extended form post support
-        opener = self._build_opener(FormPostHandler)
```
daigles: removing dead code

shotgun_api3/shotgun.py
```diff
@@ -2374,11 +2355,17 @@ def _get_attachment_upload_info(self, is_thumbnail, filename):
         if not str(upload_info).startswith("1"):
             raise ShotgunError("Could not get upload_url but " \
                                "not sure why.\nPath: %s\nUrl: %s\nError: %s" % (
-                                   path, url, str(upload_info)))
+                                   filename, url, str(upload_info)))
```
daigles: bug fix: 'path' is actually not defined in this scope.

shotgun_api3/shotgun.py
```diff
@@ -2364,7 +2344,8 @@ def _get_attachment_upload_info(self, is_thumbnail, filename):
             "filename" : filename
         }
 
-        params.update(self._auth_params())
```
daigles: Moved this to the 'send_form' function.

shotgun_api3/shotgun.py
```diff
@@ -2219,33 +2215,16 @@ def _upload_to_storage(self, entity_type, entity_id, path, field_name, display_n
 
         # Step 1: get the upload url
 
-        upload_info = self._get_attachment_upload_info(is_thumbnail, filename)
+        multipart_upload = (os.path.getsize(path)  > 20000000)
```
manneohrstrom: nit: on the toolkit side we would probably prefix with is to indicate boolean, e.g. `is_multipart_upload`

shotgun_api3/shotgun.py
```diff
@@ -2219,33 +2215,16 @@ def _upload_to_storage(self, entity_type, entity_id, path, field_name, display_n
...
-                raise ShotgunError("Server encountered an internal error.\n%s\n%s\n\n" % (url, e))
-            else:
-                raise ShotgunError("Unanticipated error occurred uploading %s: %s" % (path, e))
-        finally:
-            fd.close()
+        # Step 2: upload the file
 
-        LOG.debug("File uploaded to Cloud storage: %s", filename)
+        if multipart_upload:
+            self._multipart_upload_file_to_storage(path, upload_info)
```
manneohrstrom: minor: some comments explaining why we need to make the distinction here based on file size could helpful.

shotgun_api3/shotgun.py
```diff
@@ -2364,7 +2344,8 @@ def _get_attachment_upload_info(self, is_thumbnail, filename):
             "filename" : filename
         }
 
-        params.update(self._auth_params())
+        if multipart_upload:
+            params['multipart_upload'] = True
```
manneohrstrom: double quotes pls

shotgun_api3/shotgun.py
```diff
@@ -2342,12 +2321,13 @@ def _upload_to_sg(self, entity_type, entity_id, path, field_name, display_name,
...
-    def _get_attachment_upload_info(self, is_thumbnail, filename):
+    def _get_attachment_upload_info(self, is_thumbnail, filename, multipart_upload):
         """
         Internal function to get the information needed to upload a file to Cloud storage.
 
         :param bool is_thumbnail: indicates if the attachment is a thumbnail.
         :param str filename: name of the file that will be uploaded.
+        :param bool multipart_upload: Indicates if we want multi-part upload information back.
 
         :returns: dictionary containing the upload url and
```
manneohrstrom: please update the returns docstring to reflect the new fields we are returning

shotgun_api3/shotgun.py
```diff
@@ -2342,12 +2321,13 @@ def _upload_to_sg(self, entity_type, entity_id, path, field_name, display_name,
...
         return attachment_id
 
-    def _get_attachment_upload_info(self, is_thumbnail, filename):
+    def _get_attachment_upload_info(self, is_thumbnail, filename, multipart_upload):
         """
         Internal function to get the information needed to upload a file to Cloud storage.
 
         :param bool is_thumbnail: indicates if the attachment is a thumbnail.
         :param str filename: name of the file that will be uploaded.
+        :param bool multipart_upload: Indicates if we want multi-part upload information back.
```
manneohrstrom: minor: might be worth explaining a bit more around what the difference is - does the return data change depending on this flag?

shotgun_api3/shotgun.py
```diff
@@ -2374,11 +2355,17 @@ def _get_attachment_upload_info(self, is_thumbnail, filename):
         if not str(upload_info).startswith("1"):
             raise ShotgunError("Could not get upload_url but " \
                                "not sure why.\nPath: %s\nUrl: %s\nError: %s" % (
-                                   path, url, str(upload_info)))
+                                   filename, url, str(upload_info)))
 
         LOG.debug("Completed rpc call to %s" % (upload_url))
+
+        upload_info_parts = str(upload_info).split("\n")
```
manneohrstrom: do we need to worry about different versions of the sg server responding in different ways here? or is it consistent - not sure if the server side changes are going out as a patch release or if they are perhaps already there. If they are going out in a patch, we might have to add checks to handle different server responses differently?

shotgun_api3/shotgun.py
```diff
@@ -3539,6 +3526,137 @@ def _dict_to_extra_data(self, d, key_name="value"):
...
+        try:
+            content_type = mimetypes.guess_type(path)[0]
+            content_type = content_type or "application/octet-stream"
+            file_size = os.fstat(fd.fileno())[stat.ST_SIZE]
+            filename = os.path.basename(path)
+
+            etags = []
+            part_number = 1
+            bytes_read = 0
+            chunk_size = 20000000
```
manneohrstrom: could we reuse that constant here?

shotgun_api3/shotgun.py
```diff
@@ -3539,6 +3526,137 @@ def _dict_to_extra_data(self, d, key_name="value"):
...
+            fd.close()
+
+        LOG.debug("File uploaded to Cloud storage: %s", filename)
+
+    def _multipart_upload_file_to_storage(self, path, upload_info):
+        """
+        Internal function to upload a file to the Cloud storage in multiple parts.
+
+        :param str path: Full path to an existing non-empty file on disk to upload.
+        :param dict upload_info: Contains details received from the server, about the upload.
```
manneohrstrom: would it helpful to document the expected contents of this dict, at least in one of the methods? It looks like we are accessing specific keys, so it's not just a case of sending it blindly back to the server. If we document it in one place there rest of the docstring could point to that one i guess....

shotgun_api3/shotgun.py
```diff
@@ -3539,6 +3526,137 @@ def _dict_to_extra_data(self, d, key_name="value"):
...
+            bytes_read = 0
+            chunk_size = 20000000
+            while bytes_read < file_size:
+                data = fd.read(chunk_size)
+                bytes_read += len(data)
+                part_url = self._get_upload_part_link(upload_info, filename, part_number)
+                etags.append(self._upload_data_to_storage(data, content_type, len(data), part_url ))
+                part_number += 1
+
+            self._complete_multipart_upload(upload_info, filename, etags)
```
manneohrstrom: what happens if it fails half way? Should we worry about that case? should we send a message to the server saying that things broke half way and to discard the upload? Or will the server simply discard any "dangling" pieces after a certain amount of idle time?

I think a piece of DX to consider (and point out in the code) here is that we want a user to get the same experience when a multi-part upload fails and when a normal upload fails. Is this currently the case? Would be great with some tests to assert this.

I am thinking about the specific case where clients have flaky connections (not uncommon) so one data exchange with the server may work fine but the next one fails due to external factors.

shotgun_api3/shotgun.py
```diff
@@ -3539,6 +3526,137 @@ def _dict_to_extra_data(self, d, key_name="value"):
...
+
+        :param dict upload_info: Contains details received from the server, about the upload.
+        :param str filename: Name of the file for which we want the link.
+        :param int part_number: Part number for the link.
+        :returns: upload url.
+        :rtype: str
+        """
+        params = {
+            "upload_type": upload_info["upload_type"],
+            "filename": filename,
```
manneohrstrom: curious - why is the filename important here? i am guessing that the server relies on the `upload_id` to piece things together... why do we need to send the filename as well?

might be worth adding some comments to explain

shotgun_api3/shotgun.py
```diff
@@ -3539,6 +3526,137 @@ def _dict_to_extra_data(self, d, key_name="value"):
...
+        Internal function to upload data to Cloud storage.
+        
+        :param stream data: Contains details received from the server, about the upload.
+        :param str content_type: Content type of the data stream.
+        :param int size: Number of bytes in the data stream.
+        :param str storage_url: Target URL for the uploaded file. 
+        :returns: upload url.
+        :rtype: str
+        """
+        try:
```
manneohrstrom: logging here would be great

shotgun_api3/shotgun.py
```diff
@@ -3539,6 +3526,137 @@ def _dict_to_extra_data(self, d, key_name="value"):
...
+            "timestamp": upload_info["timestamp"],
+            "upload_id": upload_info["upload_id"],
+            "etags": ",".join(etags)
+        }
+
+        url = urlparse.urlunparse((self.config.scheme, self.config.server,
+                                   "/upload/api_complete_multipart_upload", None, None, None))
+        result = self._send_form(url, params)
+
+        if not str(result).startswith("1"):
```
manneohrstrom: please add comments explaining what the response looks like

shotgun_api3/shotgun.py
daigles: done!

shotgun_api3/shotgun.py
daigles: I intentionally kept this somewhat opaque to help keep users away from the details...

shotgun_api3/shotgun.py
daigles: we're good! only 1 version of the server-side code is used!

shotgun_api3/shotgun.py
daigles: My intention is really to keep the details of the 'upload_info' as opaque as possible.
This info is used by private methods managing the upload and users should not be messing with these...

shotgun_api3/shotgun.py
daigles: server needs both the filename and upload id...

shotgun_api3/shotgun.py
manneohrstrom: sweet!

shotgun_api3/shotgun.py
manneohrstrom: makes sense!

PR #138, opened 2017-03-07
fix nested multi-entities query

renaudll:
Hi, 

We started using mockgun for testing our pipeline at Squeeze.
However I've realized that some calls to find that worked with Shotgun failed with mockgun.
This was mainly related to :
- Using nested multi-entity when filtering (ex: ["project.Project.users.HumanUser.login", "is", "user1"])
- Requested fields (indirectly through the order argument) not being returned.

I've made the changes and implemented some additional tests, hope it help!

### DIFF
--- a/shotgun_api3/lib/mockgun/mockgun.py
+++ b/shotgun_api3/lib/mockgun/mockgun.py
@@ -114,7 +114,6 @@ Below is a non-exhaustive list of things that we still need to implement:
 
 """
 
-import copy
 import datetime
 
 from ... import sg_timezone, ShotgunError
@@ -301,7 +300,7 @@ class Shotgun(object):
             )
         ]
 
-        # Request additional fields from results
+        # Extract fields from row
         val = [dict((field, self._get_field_from_row(entity_type, row, field)) for field in all_fields) for row in results]
 
         # Handle the ordering of the result after we requested additional fields from results.
@@ -379,7 +378,9 @@ class Shotgun(object):
         
         self._update_row(entity_type, row, data)
         row["id"] = next_id
-        row["created_at"] = datetime.datetime.now()
+        
+        if "create_at" not in row:
+            row["created_at"] = datetime.datetime.now()
         
         self._db[entity_type][next_id] = row
         
@@ -655,11 +656,6 @@ class Shotgun(object):
                 # not multi entity, must be entity.
                 elif not isinstance(field_value, dict):
                     raise ShotgunError("Invalid deep query field %s.%s" % (entity_type, field))
-
-                # make sure that types in the query match type in the linked field
-                if entity_type2 != field_value["type"]:
-                    raise ShotgunError("Deep query field %s.%s does not match type "
-                                       "with data %s" % (entity_type, field, field_value))
 
                 # ok so looks like the value is an entity link
                 # e.g. db contains: {"sg_sequence": {"type":"Sequence", "id": 123 } }
--- a/shotgun_api3/lib/mockgun/mockgun.py
+++ b/shotgun_api3/lib/mockgun/mockgun.py
@@ -379,7 +379,7 @@ class Shotgun(object):
         self._update_row(entity_type, row, data)
         row["id"] = next_id
         
-        if "create_at" not in row:
+        if "created_at" not in row:
             row["created_at"] = datetime.datetime.now()
         
         self._db[entity_type][next_id] = row
--- a/shotgun_api3/lib/mockgun/mockgun.py
+++ b/shotgun_api3/lib/mockgun/mockgun.py
@@ -114,6 +114,7 @@ Below is a non-exhaustive list of things that we still need to implement:
 
 """
 
+import copy
 import datetime
 
 from ... import sg_timezone, ShotgunError
@@ -242,7 +243,6 @@ class Shotgun(object):
         else:
             return dict((k, v) for k, v in self._schema[entity_type].items() if k == field_name)
 
-
     def find(self, entity_type, filters, fields=None, order=None, filter_operator=None, limit=0, retired_only=False, page=0):
         self.finds += 1
 
@@ -252,14 +252,21 @@ class Shotgun(object):
 
         # Configure fields
         if fields is None:
-            fields = set(["type", "id"])
+            requested_fields = set(["type", "id"])
         else:
-            fields = set(fields) | set(["type", "id"])
+            requested_fields = set(fields) | set(["type", "id"])
 
         # Include fields from the order argument in the searched fields
+        order_fields = set()
         if order:
             for o in order:
-                fields.add(o['field_name'])
+                order_fields.add(o['field_name'])
+
+        # Merge the requested fields and the order fields.
+        # We need all those to property order the results but still return only requested fields.
+        all_fields = set()
+        all_fields.update(requested_fields)
+        all_fields.update(order_fields)
 
         # FIXME: This should be refactored so that we can use the complex filer
         # style in nested filter operations.
@@ -283,7 +290,7 @@ class Shotgun(object):
                     resolved_filters.append([ f["path"], f["relation"], f["values"][0] ])
 
         else:
-            # traditiona style sg filters
+            # traditional style sg filters
             resolved_filters = filters
 
         results = [
@@ -295,9 +302,9 @@ class Shotgun(object):
         ]
 
         # Request additional fields from results
-        val = [dict((field, self._get_field_from_row(entity_type, row, field)) for field in fields) for row in results]
+        val = [dict((field, self._get_field_from_row(entity_type, row, field)) for field in all_fields) for row in results]
 
-        # Handle the ordering of the result after we requested additionel fields from results.
+        # Handle the ordering of the result after we requested additional fields from results.
         if order:
             # order: [{"field_name": "code", "direction": "asc"}, ... ]
             for order_entry in order:
@@ -313,6 +320,12 @@ class Shotgun(object):
                     raise ValueError("Unknown ordering direction")
 
                 val = sorted(val, key=lambda k: k[order_field], reverse=desc_order)
+
+        # Remove any fields that was not explicitely requested.
+        fields_to_remove = all_fields - requested_fields
+        for v in val:
+            for field in fields_to_remove:
+                v.pop(field)
 
         return val
 
@@ -364,8 +377,9 @@ class Shotgun(object):
         
         row = self._get_new_row(entity_type)
         
-        self._update_row(entity_type, row, data)        
+        self._update_row(entity_type, row, data)
         row["id"] = next_id
+        row["created_at"] = datetime.datetime.now()
         
         self._db[entity_type][next_id] = row
         
--- a/shotgun_api3/lib/mockgun/mockgun.py
+++ b/shotgun_api3/lib/mockgun/mockgun.py
@@ -612,25 +612,12 @@ class Shotgun(object):
     def _get_field_from_row(self, entity_type, row, field):
         # split dotted form fields
         try:
-            tokens = field.split(".")
-            # dotted form fields can have the following form:
-            # - field.EntityType.field
-            # - field.EntityType.field.EntityType.field
-            # - etc
-            num_tokens = len(tokens)
-            if num_tokens % 2 != 1:
-                raise ValueError()
+            # is it something like sg_sequence.Sequence.code ?
+            field2, entity_type2, field3 = field.split(".", 2)
 
-            current_entity = row
+            if field2 in row:
 
-            for i in range(num_tokens / 2):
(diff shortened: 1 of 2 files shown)


neilvictorgrey:
Thanks for the PR @renaudll — we'll have a couple of our people run through this over the next couple days :)

CC: @manneohrstrom @robblau

jfboismenu:
Hi @renaudll , I'm dropping by just to let you know that I will be reviewing your fix later this week. We're so happy that people are contributing to mockgun. :) Shouldn't take long.

tests/test_mockgun.py
```diff
@@ -269,6 +272,91 @@ def test_find_by_sub_entity_field(self):
...
+        """Creates test data."""
+        self._mockgun = Mockgun("https://test.shotgunstudio.com", login="user", password="1234")
+
+        # Create two users to assign to the pipeline configurations.
+        self._user1 = self._mockgun.create("HumanUser", {"login": "user1", "firstname": "firstname1"})
+
+        # Create a project for nested sub entity field
+        self._project1 = self._mockgun.create("Project", {"name": "project1", "users": [self._user1]})
+
+        # Create pipeline configurations that are assigned none, one or two users.
```
jfboismenu: Thanks for adding this test, altough looking at the code in master it appears the original code worked. :)

shotgun_api3/lib/mockgun/mockgun.py
```diff
@@ -625,31 +641,29 @@ def _get_field_from_row(self, entity_type, row, field):
...
                             continue
 
                         entity = self._db[linked_row["type"]][linked_row["id"]]
 
-                        sub_field_value = self._get_field_from_row(entity_type2, entity, field3)
+                        sub_field_value = self._get_field_from_row(entity_type, entity, rfield)
                         values.append(sub_field_value)
                     return values
                 # not multi entity, must be entity.
+                elif field_value is None:
```
jfboismenu: Adding these

```
elif field_value is None:
    return None
```

two lines to the original method are all that is necessary for your tests to pass. I recommend you restore the original code.

tests/test_mockgun.py
```diff
@@ -269,6 +272,91 @@ def test_find_by_sub_entity_field(self):
...
+        self.assertEqual(self._user2['id'], item['id'])
+
+    def test_find_order_multi_entity(self):
+        """Ensure we are able to sort data using a multi-entity field."""
+        # Test ascending order
+        item = self._mockgun.find_one("PipelineConfiguration", [], order=[{'field_name': 'users.HumanUser.login', 'direction': 'asc'}])
+        self.assertEqual(self._pipeline_configutation_2['id'], item['id'])
+
+        # Test descending order
+        item = self._mockgun.find_one("PipelineConfiguration", [], order=[{'field_name': 'users.HumanUser.login', 'direction': 'desc'}])
```
jfboismenu: I'm surprised that we could sort on a multi entity field. If there are multiple entities attached to the field which value are we supposed to even use for sorting. I'll check back with the backend gurus.

jfboismenu: To be clear, there is nothing wrong with the update to the order code you provided (except for it returning an extra field, as discussed above), so if you could just provide a unit test that would succeed on a real server that would be great. If I am indeed right that we can't order on multi-entity field values then we will probably want at a later date fix Mockgun to prevent it.

shotgun_api3/lib/mockgun/mockgun.py
```diff
@@ -244,14 +244,23 @@ def schema_field_read(self, entity_type, field_name=None):
...
 
+        # Configure fields
+        if fields is None:
+            fields = set(["type", "id"])
+        else:
+            fields = set(fields) | set(["type", "id"])
+
+        # Include fields from the order argument in the searched fields
+        if order:
+            for o in order:
```
renaudll: Your are right, the fields were leakings! I've pushed a new commit that fix this.

tests/test_mockgun.py
renaudll: We are using a lot of sorting on nested entity field on our side.
For example in production we have a custom entity named Episode that contain Sequences.
Here is the code I use to get all of it's sequence in the correct order.
```
filters = [
["sg_sequence.Sequence.sg_episode", "is", {'type': self.type, 'id': self.id}]
]
order = [
    {'field_name': 'sg_sequence.Sequence.sg_sequence_order', 'direction': 'asc'},
    {'field_name': 'sg_cut_order', 'direction': 'asc'},
]
```
However I just realised that the sg_episode and sg_sequence are 'entity' fields and the PipelineConfiguration.users is an 'multi_entity' fields. You might be right it may be not possible to sort using 'multi_entity' fields. 

I was using the 'PipelineConfigure.users' field to have a simple test case without the need to create custom fields, however I might have to rename 'test_find_order_multi_entity' to 'test_find_order_nested_entity' and design a better test case that exclusively use 'entity' fields instead of 'multi_entity' fields. I'll think of something and update the pull request accordingly!

shotgun_api3/lib/mockgun/mockgun.py
jfboismenu: Did you undo the code in `find` as well? I meant that you should undo the changes from `_get_field_from_row` method. I've pushed a branch that restores the original code from `_get_field_from_row` and injects the two lines. You can see the diff with `master` [here](https://github.com/shotgunsoftware/python-api/compare/squeeze_fixes_update?expand=1).

When running the tests on that branch, everything still passes.

tests/test_mockgun.py
jfboismenu: How about sorting pipeline configurations based on the project's name? :D

jfboismenu:
Hi @renaudll , do you have any updates?

renaudll:
Hi @jfboismenu , sorry for the delay. I want to continue working on this but don't have any free time atm. I will try sending an update next friday. Thanks!

jfboismenu:
Great! Sorry, I didn't mean to sound pushy. We appreciate the time you are taking to fix this.
Looking forward to the update.

shotgun_api3/lib/mockgun/mockgun.py
renaudll: You are also right! I've continued my work from your branch.

tests/test_mockgun.py
renaudll: Yes good idea, I've pushed an update with this change. In a perfect world I would have loved to test complex nested queries like 'field1.EntityType.field2.EntityType.field3' etc but this should do the trick for now.

renaudll:
Hi @jfboismenu , sorry for the delay, it was a crazy week! I've pushed an update that address all the points listed. Thank you so much for your time reviewing this and sorry again for the slow response. 

P.S.: I've took the liberty of adding another little fix that set the 'create_at' field on the create method (https://github.com/SqueezeStudioAnimation/python-api/commit/45ce1729304610804e39dc77ce1999fa47ce05be#diff-443157c398f7038db0d1eb0401aa7d4eR382). Don't hesitate to ask for a separated pull request!

tests/test_mockgun.py
```diff
@@ -269,6 +273,112 @@ def test_find_by_sub_entity_field(self):
...
+        self.assertEqual(self._pipeline_configutation_1['id'], item['id'])
+
+        # Test descending order
+        item = self._mockgun.find_one("PipelineConfiguration", [], order=[{'field_name': 'project.Project.name', 'direction': 'desc'}])
+        self.assertEqual(self._pipeline_configutation_2['id'], item['id'])
+
+    def test_find_order_fields_leak(self):
+        """Ensure that additional fields passed through the order argument but NOT via the field argument are not added to the resulting fields."""
+        item = self._mockgun.find_one("HumanUser", [], fields=['email'], order=[{'field_name': 'login', 'direction': 'asc'}])
+        self.assertSetEqual(set(item.keys()), set(['id', 'type', 'email']))  # note: we don't verify the order yet
```
jfboismenu: Unfortunately, `assertSetEqual` was introduced in Python 2.7 and we still need to support 2.6. `assertEqual` is our only friend for this.

shotgun_api3/lib/mockgun/mockgun.py
```diff
@@ -361,8 +377,9 @@ def create(self, entity_type, data, return_fields=None):
         
         row = self._get_new_row(entity_type)
         
-        self._update_row(entity_type, row, data)        
+        self._update_row(entity_type, row, data)
         row["id"] = next_id
+        row["created_at"] = datetime.datetime.now()
```
jfboismenu: `created_at` can be set by a `shotgun.create` call. You should test for `"created_at" in data` first.

shotgun_api3/lib/mockgun/mockgun.py
```diff
@@ -285,7 +301,10 @@ def find(self, entity_type, filters, fields=None, order=None, filter_operator=No
             )
         ]
 
-        # handle the ordering of the recordset
+        # Request additional fields from results
```
jfboismenu: I find the wording a bit confusing here, especially with `additional`. `Extract fields from rows` seems reasonable to me. Sorry to be nitpicky, but that comment feels confusing to me.

shotgun_api3/lib/mockgun/mockgun.py
```diff
@@ -300,23 +319,20 @@ def find(self, entity_type, filters, fields=None, order=None, filter_operator=No
...
-        else:
-            fields = set(fields) | set(["type", "id"])
-
-        # get the values requested
-        val = [dict((field, self._get_field_from_row(entity_type, row, field)) for field in fields) for row in results]
+        # Remove any fields that was not explicitely requested.
+        fields_to_remove = all_fields - requested_fields
+        for v in val:
+            for field in fields_to_remove:
+                v.pop(field)
```
jfboismenu: Ah, never noticed that `dict` had a `pop` method just like `list`. Looks nicer than using `del v[field]`. You get the result back as well.

Nice!

shotgun_api3/lib/mockgun/mockgun.py
```diff
@@ -114,6 +114,7 @@
 
 """
 
+import copy
```
jfboismenu: Unused.

renaudll:
Hi @jfboismenu , sorry again for the delay.

I've done the tweaks, however I still needed to modify mockgun a bit to ensure our pipeline unit-tests are working accordingly. see https://github.com/SqueezeStudioAnimation/python-api/commit/7ea4244347dfdebbd6c8e1a185cba2dd7be6c0b4#diff-443157c398f7038db0d1eb0401aa7d4eL659

I've added tests in test_find_by_sub_entity_field that demonstrate the issue. What do you think?

shotgun_api3/lib/mockgun/mockgun.py
```diff
@@ -379,7 +378,9 @@ def create(self, entity_type, data, return_fields=None):
         
         self._update_row(entity_type, row, data)
         row["id"] = next_id
-        row["created_at"] = datetime.datetime.now()
+        
+        if "create_at" not in row:
```
jfboismenu: createD_at

shotgun_api3/lib/mockgun/mockgun.py
```diff
@@ -656,11 +657,6 @@ def _get_field_from_row(self, entity_type, row, field):
                 elif not isinstance(field_value, dict):
                     raise ShotgunError("Invalid deep query field %s.%s" % (entity_type, field))
 
-                # make sure that types in the query match type in the linked field
-                if entity_type2 != field_value["type"]:
```
jfboismenu: Could you elaborate on why removing this was needed?

shotgun_api3/lib/mockgun/mockgun.py
renaudll: Fixed! Sorry about that!

shotgun_api3/lib/mockgun/mockgun.py
renaudll: You can check the test_find_by_sub_entity_field I've added in my previous commit.
It seem that this error get triggered when filtering with multi-entities. This work on a real shotgun server but fail in mockgun.
A simple example is the 'entity' field on a 'Task'.
sg.find('Task', [['entity.Asset.code', 'is', 999]])
and
sg.find('Task', [['entity.Shot.code', 'is', 999]])
Are valid queries.

If check if important for something else, I might be able to extend it to work better with multi entities by querying what are the valid entity types.

jfboismenu: The previous code was incorrect, let's remote it. We're in a better shape now.

jfboismenu:
Thank you so much @renaudll . This was awesome.

jfboismenu:
Sorry but we're going to have to reopen this one. Seems like there are a few tests failing.

PR #144, opened 2017-03-22
Fixes `is` and `is_not` comparison with `None` on multi entity fields

jfboismenu:
Title says it all. :)

### DIFF
--- a/tests/test_mockgun.py
+++ b/tests/test_mockgun.py
@@ -269,6 +269,17 @@ class TestMultiEntityFieldComparison(TestBaseWithExceptionTests):
         )
         self.assertEqual(len(items), 1)
 
+    def test_find_with_none(self):
+
+        items = self._mockgun.find("PipelineConfiguration", [["users", "is", None]], ["users"])
+        self.assertEqual(len(items), 1)
+        self.assertListEqual(items[0]["users"], [])
+
+        items = self._mockgun.find("PipelineConfiguration", [["users", "is_not", None]], ["users"])
+        self.assertEqual(len(items), 3)
+        for item in items:
+            self.assertTrue(len(item["users"]) > 0)
+
 
 class TestFilterOperator(TestBaseWithExceptionTests):
     """
--- a/tests/test_mockgun.py
+++ b/tests/test_mockgun.py
@@ -275,7 +275,7 @@ class TestMultiEntityFieldComparison(TestBaseWithExceptionTests):
         """
         items = self._mockgun.find("PipelineConfiguration", [["users", "is", None]], ["users"])
         self.assertEqual(len(items), 1)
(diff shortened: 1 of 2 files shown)


tests/test_mockgun.py
```diff
@@ -269,6 +269,19 @@ def test_find_by_sub_entity_field(self):
         )
         self.assertEqual(len(items), 1)
 
+    def test_find_with_none(self):
+        """
+        Ensures comparison with multi-entity fields and None works.
+        """
+        items = self._mockgun.find("PipelineConfiguration", [["users", "is", None]], ["users"])
+        self.assertEqual(len(items), 1)
```
foutoucour: Minor: By convention, asserEqual first argument is the expected value, the second would be the value that is unknown.

tests/test_mockgun.py
```diff
@@ -269,6 +269,19 @@ def test_find_by_sub_entity_field(self):
...
+        self.assertEqual(items[0]["users"], [])
+
+        items = self._mockgun.find("PipelineConfiguration", [["users", "is_not", None]], ["users"])
+        self.assertEqual(len(items), 3)
+        for item in items:
```
foutoucour: Minor: As there is not message that details the test, I would go for a single `self.assertTrue(all(len(item("users"]) > 0 for item in items, msg="All the queried items should be have no users.")`

It should speed up the test and increase readability

tests/test_mockgun.py
```diff
@@ -269,6 +269,19 @@ def test_find_by_sub_entity_field(self):
...
 
+    def test_find_with_none(self):
+        """
+        Ensures comparison with multi-entity fields and None works.
+        """
+        items = self._mockgun.find("PipelineConfiguration", [["users", "is", None]], ["users"])
+        self.assertEqual(len(items), 1)
+        self.assertEqual(items[0]["users"], [])
+
+        items = self._mockgun.find("PipelineConfiguration", [["users", "is_not", None]], ["users"])
```
foutoucour: This would deserve a second test. One for the case of `is` and a second for `is_not`.

It will increase the readability of the tests

jfboismenu:
Hi @foutoucour , I love that you are jumping on a PR. Don't hesitate to do that. :) This is much appreciated.

This has already been merged in, but I'll leave it open so it's clear that something still needs to be addressed.

Thanks!

foutoucour:
@jfboismenu everything is pretty minor here. I would close the PR. 

At last, I would open an issue for tracking, but even then, it is more noise than anything.

jfboismenu:
Was merged in https://github.com/shotgunsoftware/python-api/commit/2a8b7ce8b93e7e79583499857fd575a7c1fe62b8

PR #145, opened 2017-03-30
mockgun: review _update_row to add the case of date fields

foutoucour:
* The official python API returns a string instead of the date object.
    This patch mimics the behavior
    Keep in mind the python API returns a datetime object in case of
    datetimes.
* Add docstring
* Set the cases outside to the logic.

### DIFF
--- a/shotgun_api3/lib/mockgun/mockgun.py
+++ b/shotgun_api3/lib/mockgun/mockgun.py
@@ -773,8 +773,8 @@ class Shotgun(object):
         :param dict data: data to inject in the row.
         """
         cases = {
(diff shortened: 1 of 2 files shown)


foutoucour:
@jfboismenu hey dude. Any guide on how to do/run the tests for mockgun?

jfboismenu:
@foutoucour `python test_mockgun.py` :)

foutoucour:
I changed the implementation as the issue wasn't actually in `_update_row` but in the validation of the field.
The `date` should be a string in the database but provided as a string too.

Added tests too!

foutoucour:
@jfboismenu should I close this branch?

PR #146, opened 2017-03-30
For #41856: added function to support new nav_search RPC

daigles:


HISTORY.rst
```diff
@@ -18,6 +18,7 @@ v3.0.33.dev (TBD)
    - Added unit tests for Mockgun in `test_mockgun.py`. The tests currently cover all the fixes and features mentionned above.
    - Schemas are written using pickle's highest protocol as of Python 2.4's release, which speeds up the loading from disk.
    - The current schema is cached in memory so it can be reused without having to do back to disk.
+- Added ``nav_search_string()`` and ``nav_search_entity()`` methods as an experimental, internal method for querying SG hierarchy.
```
manneohrstrom: nit: as experimental, internal methods for querying...

shotgun_api3/shotgun.py
```diff
@@ -2895,6 +2895,43 @@ def nav_expand(self, path, seed_entity_field=None, entity_fields=None):
             }
         )
 
+    def nav_search_string(self, path_filter, search_string):
```
manneohrstrom: this is probably my fault because i was using `path_filter` when we were chatting - sorry! Seeing it alongside `def nav_expand(self, path, seed_entity_field=None, entity_fields=None)` i now wonder if it's better to call it 
`def nav_search_string(self, path, search_string):` for consistency. What do you think? `root_path` is another suggestion to emphasize that we are requesting a search  below the specified path tree

shotgun_api3/shotgun.py
```diff
@@ -2895,6 +2895,43 @@ def nav_expand(self, path, seed_entity_field=None, entity_fields=None):
...
+        Search function adapted to work with the navigation hierarchy.
+
+        .. warning::
+
+            This is an experimental method that is not officially part of the
+            python-api. Usage of this method is discouraged. This method's name,
+            arguments, and argument types may change at any point.
+        """
+        return self._call_rpc(
+                'nav_search',
```
manneohrstrom: double quotes for all these guys please ;)

daigles: :\

shotgun_api3/shotgun.py
daigles: I didn't like this name. I'll go with root_path!

PR #148, opened 2017-06-24
For #43466 - Pick up records_per_page from the server capabilities

pboucher:


daigles:
There is a 'server_info' method (https://github.com/shotgunsoftware/python-api/blob/master/shotgun_api3/shotgun.py#L589) 
Should this be used instead? And update the doc around this function to indicate this new field could be present?

PR #149, opened 2017-07-10
Add support for external SSL certificates in upload methods.

mike1158:
Note: Have tested that this works at our local site, but haven't tested it with proxy connections.


jfboismenu:
Hey Mike, we've just merged in an [alternative implementation](https://github.com/shotgunsoftware/python-api/commit/8c2584b150e931624f1c66ab5bc4e26a1cd81155), which is backwards compatible with Python 2.6. It's not officially released yet, but we'd love to hear from you and see if it works on your end.

PR #151, opened 2017-08-01
Fix httplib2 integration

### LINKED ISSUE #150: Bundle httplib2 should be adapted for the integration you've made of it
Hi,

noticed with this stack trace : 

```
 File "../python-2.7/shotgun_api3/sg_26.py", line 5, in <module>
    from .lib.httplib2 import Http, ProxyInfo, socks, SSLHandshakeError
 File "../python-2.7/shotgun_api3/lib/httplib2/__init__.py", line 59, in <module>
    from httplib2 import socks
  File "../python-2.7/httplib2/__init__.py", line 915, in <module>
    class HTTPSConnectionWithTimeout(httplib.HTTPSConnection):
AttributeError: 'module' object has no attribute 'HTTPSConnection'
```

As you can see : 

you have **_integrated_** `httplib2` into shotgun_api3.lib _sub-package_.

you correctly use it with it :     `from .lib.httplib2 import Http, ProxyInfo, socks, SSLHandshakeError`

but then you should also have adapted the import at line 59 of `shotgun_api3/lib/httplib2/__init__.py`  to `from shotgun_api3.lib.httplib2 import socks` or the better `from . import socks`.

cause like that you actually use a part of the integrated httplib2 you made a copy and a part of the _normal_ httplib2 which can comes in any environment where we need to use shotgun_api3.. which may or may not be compatible..

I don't know if there are no other such import to adapt in the different things you have integrated into/under shotgun_api3/lib/*

gst:
should fix #150

### DIFF
--- a/shotgun_api3/lib/httplib2/__init__.py
+++ b/shotgun_api3/lib/httplib2/__init__.py
@@ -56,7 +56,7 @@ from gettext import gettext as _
 import socket
 
 try:
-    from httplib2 import socks
+    from . import socks
 except ImportError:
     try:
         import socks


carlos-villavicencio-adsk:
Fixed on #202

PR #165, opened 2018-02-14
Added best_practices.py file to the docs/examples folder.

raphaelmatto:
An example file used in a Shotgun Developer learning video.


robblau:
Talked with @raphaelmatto and this code will move to a better location where it can be more strongly associated with the video it comes from.  Closing PR.

PR #168, opened 2018-03-16
For #46997, fix connect false

jfboismenu:
Fixes an issue where we would connect to the server at `__init__` time to retrieve the number of records per page, even if `connect=False` was specified. We will now lazily initialize that value.

The fix is designed in such a way that we are backwards compatible.
- From the library or from client code, `sg.config.records_per_page` will always return the right value from the server, if available
- Assigning `sg.config.records_per_page` will override whatever value was on the server.


shotgun_api3/shotgun.py
```diff
@@ -371,6 +375,21 @@ def __init__(self):
...
+        """
+        if self._records_per_page is None:
+            # Check for api_max_entities_per_page in the server info and change the record per page
+            # value if it is supplied.
+            self._records_per_page = self._sg.server_info.get('api_max_entities_per_page') or 500
+        return self._records_per_page
+
+    @records_per_page.setter
+    def records_per_page(self, records_per_page):
+        self._records_per_page = records_per_page
```
brandonvfx: Should we make sure someone doesn't set the value higher than `api_max_entities_per_page`?

PR #171, opened 2018-03-21
For #46920, custom CA certs for file transfer

jfboismenu:
This allows to upload and download files from local Shotgun sites using HTTPS with a custom certificate authority certificate.


shotgun_api3/shotgun.py
```diff
@@ -3034,11 +3035,15 @@ def _build_opener(self, handler):
         """
         Build urllib2 opener with appropriate proxy handler.
         """
+        handlers = []
+        if self.__ca_certs and not NO_SSL_VALIDATION:
```
jfboismenu: This fix is only applied if SHOTGUN_API_CACERTS is set, which makes it very safe as it doesn't impact any other usage of the API. The build_opener is strictly used for uploads/downloads.

jfboismenu:
All this has been tested with proxies and python 2.6

PR #172, opened 2018-03-23
For #46904, site wide tk-config-default2

jfboismenu:
While writing unit tests in Toolkit I stumbled upon this bug, where searching for a sub-entity field would crash if any of the sub-entity field was set to `None`. See unit test for more info.

Also, PEP8ed `mockgun.py`. The actual fix [is on line 679](https://github.com/shotgunsoftware/python-api/pull/172/files#diff-443157c398f7038db0d1eb0401aa7d4eR679)

### DIFF
--- a/shotgun_api3/lib/mockgun/mockgun.py
+++ b/shotgun_api3/lib/mockgun/mockgun.py
@@ -116,7 +116,7 @@ Below is a non-exhaustive list of things that we still need to implement:
 
 import datetime
 
-from ... import sg_timezone, ShotgunError
+from ... import ShotgunError
 from ...shotgun import _Config
 from .errors import MockgunError
 from .schema import SchemaFactory
@@ -220,7 +220,6 @@ class Shotgun(object):
     ###################################################################################################
     # public API methods
 
-
     def get_session_token(self):
         return "bogus_session_token"
 
@@ -245,15 +244,17 @@ class Shotgun(object):
         else:
             return dict((k, v) for k, v in self._schema[entity_type].items() if k == field_name)
 
-
-    def find(self, entity_type, filters, fields=None, order=None, filter_operator=None, limit=0, retired_only=False, page=0):
-
+    def find(
+        self,
+        entity_type, filters, fields=None,
+        order=None, filter_operator=None, limit=0, retired_only=False, page=0
+    ):
 
         self.finds += 1
 
         self._validate_entity_type(entity_type)
         # do not validate custom fields - this makes it hard to mock up a field quickly
-        #self._validate_entity_fields(entity_type, fields)
+        # self._validate_entity_fields(entity_type, fields)
 
         # FIXME: This should be refactored so that we can use the complex filer
         # style in nested filter operations.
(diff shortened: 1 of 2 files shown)


shotgun_api3/lib/mockgun/mockgun.py
```diff
@@ -245,15 +244,17 @@ def schema_field_read(self, entity_type, field_name=None):
         else:
             return dict((k, v) for k, v in self._schema[entity_type].items() if k == field_name)
 
-
-    def find(self, entity_type, filters, fields=None, order=None, filter_operator=None, limit=0, retired_only=False, page=0):
-
+    def find(
+        self,
+        entity_type, filters, fields=None,
```
josh-t: this formatting seems a bit arbitrary. Any reason it's done this way?

shotgun_api3/lib/mockgun/mockgun.py
```diff
@@ -315,9 +316,15 @@ def find(self, entity_type, filters, fields=None, order=None, filter_operator=No
 
         return val
 
-
-    def find_one(self, entity_type, filters, fields=None, order=None, filter_operator=None, retired_only=False):
-        results = self.find(entity_type, filters, fields=fields, order=order, filter_operator=filter_operator, retired_only=retired_only)
+    def find_one(
+        self,
+        entity_type, filters, fields=None,
```
josh-t: same.

shotgun_api3/lib/mockgun/mockgun.py
```diff
@@ -440,22 +447,44 @@ def _validate_entity_data(self, entity_type, data):
...
+                    )
                 elif "id" not in item or "type" not in item:
-                    raise ShotgunError("%s.%s is of type entity, but data %s does not contain 'type' and 'id'" % (entity_type, field, item))
-                #elif item["type"] not in field_info["properties"]["valid_types"]["value"]:
-                #    raise ShotgunError("%s.%s is of type entity, but data %s has an invalid type (expected one of %s)" % (entity_type, field, item, field_info["properties"]["valid_types"]["value"]))
+                    raise ShotgunError(
+                        "%s.%s is of type entity, but data %s does not contain 'type' and 'id'"
+                        % (entity_type, field, item)
+                    )
+                # elif item["type"] not in field_info["properties"]["valid_types"]["value"]:
```
josh-t: any reason why this is commented out? maybe make a comment to say why it has been left in?

shotgun_api3/lib/mockgun/mockgun.py
jfboismenu: First line is self, next three is the most common-only used params to the method and then the rest. Should I simply break them on two lines? One arg per line seems wasteful to me.

shotgun_api3/lib/mockgun/mockgun.py
jfboismenu: Maybe it's not obvious from the diff, but it was already commented out. I only reformatted it because it was longer than 120 chars. Unfortunately we don't know why it was commented out...

Issue #154, opened 2017-09-03
Add Python 3 support


mjnowen:
http://www.vfxplatform.com/

29th April 2017 - CY2018 Draft published with significant upgrades to gcc and glibc. This also comes with notice of our intention to move to Python 3 for CY2019 in advance of Python 2 support ending in 2020. We are currently soliciting feedback on this Draft so please either send to feedback@vfxplatform.com or share on vfx-platform-discuss. We expect to publish CY2018 Final around SIGGRAPH in July.

PR #174, opened 2018-05-08
py3 support

rane-hs:
#154

## auto formattting
 using autopep8 : https://github.com/hhatto/autopep8

## breaking backward compatibility : python2
 execute `2to3 -w ./` converting whole file to changing py3 implementation

## breaking backward compatibility : choose_boundary
chose_boundary is no longer available in python3 
https://stackoverflow.com/questions/27099290/where-is-mimetools-choose-boundary-function-in-python3?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa
switching `make_boundary`

or using converted function https://gist.github.com/rane-hs/40b36be270797fe23480885341121d85


thebeeland:
Thanks for sending this to us, @rane-hs, but we're not going to be able to merge this in as things stand right now. We're planning to align with the VFX Platform schedule for supporting Python 3.x. As that timeline progresses, we'll be offering the same kinds of developer previews as are planned for other VFX Platform applications. You can keep an eye on the shotgun-dev email list in the meantime, as any updates regarding specific dates will be posted there when we know them.

PR #179, opened 2018-06-21
SG-4483: Handles non-ascii unicode string paths gracefully during upload.

thebeeland:
If a unicode string that couldn't be encoded as ascii was given to the upload method, it would cause an unhandled UnicodeEncodeError to be raised. We now detect the unicode string paths and attempt to encode them as utf-8 strings before uploading.


shotgun_api3/shotgun.py
```diff
@@ -2365,6 +2365,20 @@ def _upload_to_sg(self, entity_type, entity_id, path, field_name, display_name,
...
 
+        # If we ended up with a unicode string path, we need to encode it
+        # as a utf-8 string. If we don't, there's a chance that there will
+        # will be an attempt later on to encode it as an ascii string, and
+        # that will fail ungracefully if the path contains any non-ascii
+        # characters.
+        if isinstance(path, unicode):
+            try:
+                path = path.encode("utf-8")
+            except UnicodeEncodeError:
```
manneohrstrom: I might be wrong, but I don't think we need this exception check. All unicode characters can be turned into utf-8 (_UTF-8 is a variable width character encoding capable of encoding all 1,112,064 valid code points in Unicode using one to four 8-bit bytes._ - see https://en.wikipedia.org/wiki/UTF-8)

thebeeland: Cool. I was doing it out of an over abundance of caution, but if it's completely unnecessary then I'll remove it!

shotgun_api3/shotgun.py
```diff
@@ -2365,10 +2365,27 @@ def _upload_to_sg(self, entity_type, entity_id, path, field_name, display_name,
...
+        # that will fail ungracefully if the path contains any non-ascii
+        # characters.
+        #
+        # On Windows, if the path contains non-ascii characters, the calls
+        # to open later in this method will fail to find the file if given
+        # a utf-8 encoded string path. In that case, we're going to have
+        # to call open on the unicode path, but we'll use the utf-8 encoded
+        # string for everything else.
+        path_to_open = path
+        if isinstance(path, unicode):
```
manneohrstrom: i am guessing if someone now uploads a path on windows and passes a utf-8 encoded string to the API, it will still fail, right? Maybe we can add a test for that and add a little bit more logic. Would be great if we tested in our unit tests upload with ascii string, utf-8 encoded string (with special characters), unicode only containing 8 bit ascii chars and unicode containing special characters). If those tests pass we should be ok (i think!)

manneohrstrom: on a side note, what happens if you pass a `shift-jis` encoded string to the API - might be worth adding a test for that too. This is not uncommon as a locale still in japan for example. Would be nice to validate that if the input is not utf-8 and is not unicode, we provide a sensible error message.

thebeeland: Good things to check. Thinking about it, you're probably right about getting a utf-8 encoded string on Windows and it failing.

shotgun_api3/shotgun.py
```diff
@@ -2251,6 +2251,23 @@ def upload(self, entity_type, entity_id, path, field_name=None, display_name=Non
...
+        # encoded strings, but will fail on some others if the string includes non
+        # ascii characters.
+        if not isinstance(path, unicode):
+            try:
+                path = path.decode("utf-8")
+            except UnicodeDecodeError:
+                raise ShotgunError(
+                    "Could not upload the given file path. It is encoded as "
+                    "something other than utf-8 or ascii. To upload this file, "
+                    "it can be encoded as utf-8, or given as unicode: %s" % path
```
manneohrstrom: nitpick: To be extra clear, i suggest we say "it can be a string ecoded as utf-8, or given as unicode"

tests/test_api.py
```diff
@@ -218,6 +219,63 @@ def test_upload_download(self):
...
+        # If this is a problem, it'll raise with a UnicodeEncodeError. We
+        # don't need to check the results of the upload itself -- we're
+        # only checking that the non-ascii string encoding doesn't trip
+        # us up the way it used to.
+        self.sg.upload(
+            "Ticket",
+            self.ticket['id'],
+            u_path,
+            'attachments',
+            tag_list="monkeys, everywhere, send, help"
```
manneohrstrom: 😍

shotgun_api3/shotgun.py
thebeeland: 👍

PR #181, opened 2018-07-09
SG-6892 - Use s3_enabled_upload_types for cloud uploads

mferderber:
Makes use of the new `s3_enabled_upload_types` info object to see if an upload supports and should be sent to cloud storage.

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -3856,17 +3856,41 @@ class Shotgun(object):
         Internal function that determines if an entity_type + field_name combination
         should be uploaded to cloud storage.
 
+        The info endpoint should return `s3_enabled_upload_types` which contains an object like the following:
+            {
+                'Version': ['sg_uploaded_movie'],
+                'Attachment': '*',
+                '*': ['this_file']
+            }
+
         :param str entity_type: The entity type of the file being uploaded.
         :param str field_name: The matching field name for the file being uploaded.
 
         :returns: Whether the field + entity type combination should be uploaded to cloud storage.
         :rtype: bool
         """
-        supported_s3_types = self.server_info.get('s3_enabled_upload_types', {})
-        supported_fields = supported_s3_types.get(entity_type, [])
-        return self.server_info.get("s3_direct_uploads_enabled", False) and \
-            (field_name in supported_fields or field_name in supported_s3_types.get("*", []) or
-                (isinstance(supported_fields, list) and "*" in supported_fields) or supported_fields == "*")
+        supported_s3_types = self.server_info.get('s3_enabled_upload_types') or {}
+        supported_fields = supported_s3_types.get(entity_type) or []
+        supported_star_fields = supported_s3_types.get("*") or []
+        # If direct uploads are enabled
+        if self.server_info.get("s3_direct_uploads_enabled", False):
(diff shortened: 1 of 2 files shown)


shotgun_api3/shotgun.py
```diff
@@ -3853,6 +3851,23 @@ def _complete_multipart_upload(self, upload_info, filename, etags):
...
+        should be uploaded to cloud storage.
+
+        :param str entity_type: The entity type of the file being uploaded.
+        :param str field_name: The matching field name for the file being uploaded.
+
+        :returns: Whether the field + entity type combination should be uploaded to cloud storage.
+        :rtype: bool
+        """
+        supported_s3_types = self.server_info.get('s3_enabled_upload_types', {})
+        supported_fields = supported_s3_types.get(entity_type, [])
```
manneohrstrom: nitpick: better and more idiomatic to do `supported_fields = supported_s3_types.get(entity_type) or []`

shotgun_api3/shotgun.py
```diff
@@ -3853,6 +3851,23 @@ def _complete_multipart_upload(self, upload_info, filename, etags):
...
+        return self.server_info.get("s3_direct_uploads_enabled", False) and \
+            (field_name in supported_fields or field_name in supported_s3_types.get("*", []) or
+                (isinstance(supported_fields, list) and "*" in supported_fields) or supported_fields == "*")
```
manneohrstrom: do we need `isinstance(supported_fields, list)`? Isn't it always a list?

shotgun_api3/shotgun.py
```diff
@@ -3853,6 +3851,23 @@ def _complete_multipart_upload(self, upload_info, filename, etags):
...
+        Internal function that determines if an entity_type + field_name combination
+        should be uploaded to cloud storage.
+
+        :param str entity_type: The entity type of the file being uploaded.
+        :param str field_name: The matching field name for the file being uploaded.
+
+        :returns: Whether the field + entity type combination should be uploaded to cloud storage.
+        :rtype: bool
+        """
+        supported_s3_types = self.server_info.get('s3_enabled_upload_types', {})
```
manneohrstrom: @daigles - just double checking - this code currently doesn't do any shotgun server version checking - are we safe to assume that all versions of shotgun is returning the same data structure for `s3_enabled_upload_types`?

manneohrstrom: @daigles - are we ok to use `s3` here in our code and flags? IIRC, the code that you originally put in was trying to avoid mentioning what backend storage solution was being used (since this is an implementation detail)! Now it looks like we are starting to put it in. Would be nice if we could continue to keep the code cloud-agnostic.

daigles: Good point. This code needs to work with previous versions of SG backend... The default value should correspond with what we currently support (e.g. Version, sg_uploaded_movie).

shotgun_api3/shotgun.py
mferderber: The one case I figured it wouldn't be a list is if star is the value. For example I think it makes sense for `s3_enabled_upload_types` to be:
```json
{ "Version": "*" }
```

I didn't want to make too many assumptions about what the api returned. Would it make more sense to change it to only lists to be consistent?

jmmistrot:
We still

PR #182, opened 2018-07-26
Use ProxyHandler for uploads when using proxy elsewhere

apetrynet:
To avoid having to use the `http_proxy` environment variable when uploading files behind a proxy, this patch creates a ProxyHandler rather than a HTTPHandler when the `self.config.raw_http_proxy` is set.

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -3811,9 +3811,7 @@ class Shotgun(object):
         """
         try:
             if self.config.raw_http_proxy is not None:
-                handler = urllib2.ProxyHandler(
-                              {'https': self.config.raw_http_proxy}
-                          )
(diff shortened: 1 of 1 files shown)


apetrynet:
Hmm. Not quite sure what's causing the failed tests here. Does it have to do with my change at all or is it because master (which I forked from) is broken at the moment?

neilvictorgrey:
Hi @apetrynet ,

Thanks for the Pull Request! We'll have a run through it and get back to you.

As for the failed tests, that's probably just because of our CI setup (it's not very friendly to submitters external to the shotgun org). So don't worry about that for now :).

Cheers,
Neil

apetrynet:
Cool!

apetrynet:
Seems like something similar to this is implemented, so closing the PR

PR #187, opened 2019-01-21
For SG-9540, upgrade httplib2

jfboismenu:
This upgrades httplib2 to the most recent version (0.12.0). It has been slightly modified so the library imports parts of itself via `from . import <something>` instead of `from http2lib import something`, as the library isn't part of the `PYTHONPATH`.

The reasons for this are two-fold:

1) first, it makes the API compatible with the Foundry's Athera pipeline
2) it packages a much more recent version of the cacerts.txt file.

For the cacerts, we will have to revisit how we update the cacerts file so we don't end up with an out of date list in the future.

I also refactored the setUp code from the base test classes. Since these methods were essentially loading data from Shotgun in read-only, I've moved their initialization code from `setUp` to `setUpClass`, which means these values will be read from Shotgun once for every class instead of once per test. The execution time dropped from 287 seconds to 165 seconds on my computer.


README.md
```diff
@@ -36,6 +36,12 @@ Some useful direct links:
 
 You can see the [full history of the Python API on the documentation site](http://developer.shotgunsoftware.com/python-api/changelog.html).
 
+## Updating HTTPLib2
```
manneohrstrom: nice!

PR #194, opened 2019-04-30
Ticket/sg 12185 cookbook pipeline steps update

plongitudes:
per jira ticket: client wrote in thinking that they couldn't use `step_xx` in a filter based on the current wording in the cookbook, when the meaning was just that querying the step directly will result in {{None}}. Offering a change to the description to help clarify.


plongitudes:
weird, picked up some commits that i didn't want. closing and trying again.

PR #196, opened 2019-05-01
SG-11528 Add optional sleep between retries

willis102:
Add optional sleep between request attempts defaulting to 3 seconds, and allow override through optional environment variable.


shotgun_api3/shotgun.py
```diff
@@ -345,6 +345,10 @@ def __init__(self, sg):
         """
         self._sg = sg
         self.max_rpc_attempts = 3
+        # The interval between _call_rpc retries.  Since the Config can't be
+        # modified until the Shotgun instance is created, use an environment
+        # variable to override the default value.
+        self.rpc_attempt_interval = os.environ.get("SHOTGUN_API_RETRY_INTERVAL", 3)
```
thebeeland: Could this also be an optional argument when a `Shotgun` instance is initialized? I'm thinking it would work like the `ca_certs` argument, where it can be passed in OR set via an env var. You can store the retry sleep as a property on the `Shotgun` object and `_Config` can read it from there.

That has the added benefit of giving us a good place to document this and the environment variable.

thebeeland: Also, however this is documented, it would be good to specifically mention that it's expecting to receive a value in seconds.

jfboismenu: I think there’s a bug with your env var. If it is set, you’ll get a string back, not an int. You’ll need to convert the value before using it 

Maybe I suggest having the evr var be in milliseconds instead of seconds? There sort of time settings are usually in milliseconds not seconds.

willis102: I added a kwarg to Shotgun's __init__, and I'm now storing the value on the _Config instance.  I kept the rpc_attempt_interval property on _Config, since this seemed consistent with other similar properties (notably max_rpc_attempts), but I'm happy to move it to the Shotgun object if you think that would be cleaner.

Added a cast to int when getting the env var value, and am now storing wait in ms.

shotgun_api3/shotgun.py
```diff
@@ -3346,10 +3358,15 @@ def _make_call(self, verb, path, body, headers):
                 if attempt == max_rpc_attempts:
                     raise
             except Exception:
-                #TODO: LOG ?
```
jfboismenu: It's not your code, but I'm really worried about this `except` clause. Any errors raises by the code, even programming errors on our part, will retry the request. This seems dangerous to me.

@thebeeland , should we log a tech debt ticket about this and figure out how to properly fix it? We'd have to take into account all the different http and socket error types that can be encountered here.

tests/test_client.py
```diff
@@ -225,15 +226,56 @@ def test_connect_close(self):
...
+            calls = [mock.callargs(((attempt_interval,), {}))]
+            calls *= (self.sg.config.max_rpc_attempts - 1)
+            self.assertTrue(
+                mock_sleep.call_args_list == calls,
+                "Call is repeated at correct interval."
+            )
+
+    def test_set_retry_interval(self):
+        """Setting the retry interval through parameter and environment variable works."""
+        original_env_val = os.environ.pop("SHOTGUN_API_RETRY_INTERVAL", None)
```
jfboismenu: If your test fails then the environment variable state will leak into other tests.

Maybe you could pass the value the environment variable should be set to inside `run_internal_test` and inside `run_interval_test` have something like
```
old_value =  os.environ.pop("SHOTGUN_API_RETRY_INTERVAL")
os.environ["SHOTGUN_API_RETRY_INTERVAL"] = new_retry_interval
try:
   # your test code goes here.
finally:
   if old_value is not None:
      os.environ["SHOTGUN_API_RETRY_INTERVAL"] = old_value
```

shotgun_api3/shotgun.py
thebeeland: We should, and I will. Thanks for the heads up.

shotgun_api3/shotgun.py
```diff
@@ -424,7 +426,8 @@ def __init__(self,
                  password=None,
                  sudo_as_login=None,
                  session_token=None,
-                 auth_token=None):
+                 auth_token=None,
+                 rpc_attempt_interval=None):
```
manneohrstrom: the constructor already has a gazillion paramters. We should not add any more unless they are considered high frequency use ones. Can this be a parameter instead? e.g.

```
sg = Shotgun(site_name, script_name, script_key)
sg.rpc_attempt_interval = 1000 # adjusting default interval
```

Unless we have a strong reason for using the constructor approach, i think we should go with the approach above. Thoughts?

jfboismenu: My main worry with a parameter on the instance is that it goes against the grain of the current api. I'm not saying passing in a parameter is necessarily the better approach either to be honest.

As long as it can be documented properly, I think setting it as a parameter on the instance is fine. It's not a long lasting value like the host name or the credentials.

jfboismenu: I think the only change I'd make to your proposal Manne is to make it so that
`sg.config.rpc_attempt_internal` is there the value gets set, as the config object is where we keep settings.

manneohrstrom: The pattern exists already - see for example:
https://groups.google.com/a/shotgunsoftware.com/forum/#!msg/shotgun-dev/P-YXTWs9mp0/tzXtkIALt1sJ

Yes yes! I totally agree that we should keep it on the `config` object!

willis102: I've moved `rpc_attempt_interval` back to being just a property of `_Config`, and improved the comment there, after discussing with @manneohrstrom and @thebeeland how best to document this.

tests/test_client.py
willis102: ✅wrapped in try-finally, and also ensure we won't try to pop from os.environ if the key doesn't exist.

PR #203, opened 2019-07-12
Set up CI with Azure Pipelines

jfboismenu:
Adds support for Azure Pipelines.


jfboismenu:
Closing the PR as this was meant only for a hackathon. We'll see if we want to merge this later.

PR #204, opened 2019-07-30
SG-12945: Added localized param to Shotgun object

palisir:
The goal of [SG-12945](https://jira.autodesk.com/browse/SG-12945) is to expose a new param "localised" on the object Shotgun, to give the ability to get localized fields when possible (only DisplayColumn at the time).
Documentation is on its way too.

As discussed with @thebeeland, the requirement for tests will be evaluated by the reviewer.


shotgun_api3/shotgun.py
```diff
@@ -469,7 +470,8 @@ def __init__(self,
                  password=None,
                  sudo_as_login=None,
                  session_token=None,
-                 auth_token=None):
+                 auth_token=None,
+                 localized=True):
```
thebeeland: I might default this to `False`. My fear would be that this will change the form of the data returned by the API and might break existing client scripts (or even stuff in Toolkit that we provide) without additional code changes to compensate for it.

thebeeland:
@palisir -- for testing, would it be possible to use a similar approach to the below to test that the new init argument triggers the header value to be added to the request if it's True, and that it doesn't when it's False?

https://github.com/shotgunsoftware/python-api/blob/master/tests/test_client.py#L167-L231

manneohrstrom:
Is is necessary to pass this in as part of the constructor? In other advanced cases like this (e.g. `sg.config.rpc_attempt_interval = 1000`), we don't always provide a ctor parameter - the `__init__` is bloated as it is already.

In fact, there was a similar discussion recently - see https://github.com/shotgunsoftware/python-api/pull/196 - where we ended up adding it to the config object only.

HISTORY.rst
```diff
@@ -4,6 +4,10 @@ Shotgun Python API Changelog
 
 Here you can see the full list of changes between each Python API release.
 
+v3.1.1 (2019 August 29)
```
manneohrstrom: How does this relate to this particular pull request? I am seeing a bunch of changes in the diff here that seem unrelated to this PR. Do you need to merge with master?

shotgun_api3/shotgun.py
```diff
@@ -538,6 +539,9 @@ def __init__(self,
                 ``auth_token`` is invalid.
             .. todo: Add this info to the Authentication section of the docs
 
+        :param bool localized: A boolean to asks for some fields to be localized. When ``True``, a
```
manneohrstrom: This does not match the parameters of the constructor.

shotgun_api3/shotgun.py
```diff
@@ -427,6 +427,7 @@ def __init__(self, sg):
         self.session_token = None
         self.authorization = None
         self.no_ssl_validation = False
+        self.localized = False
```
manneohrstrom: i am not seeing any docs or comments to describe what this is. Please add at least a comment - ideally sphinx docs in context of the code, but it's also ok if these are separate and part of the rst file.

palisir:
As discussed with @manneohrstrom I am closing this PR to make a new one on another branch

PR #208, opened 2019-09-05
SG-12945 + SG-12581 : Added localized param to Shotgun object

palisir:
The goal of SG-12945 is to expose a new config property named "localized" on the object Shotgun, to give the ability to get localized display names from the server when possible.

The code is tested and documented (according to SG-12581).


manneohrstrom:
This is looking awesome! Thank you, @palisir! For reference, the built docs look like this:

image

docs/reference.rst
```diff
@@ -891,3 +891,45 @@ Stores the number of milliseconds to wait between request retries.  By default,
 
 In the case that both this environment variable and the config's ``rpc_attempt_interval`` property are set, the value in ``rpc_attempt_interal`` will be used.
 
+************
+Localization
+************
+
+Shotgun API offers the ability to return display namnes in user's language.
```
manneohrstrom: *The* Shotgun API

manneohrstrom: How does the API know what the user's language is? I think this is worth explaining. Is it based on the user's language preferences in Shotgun? (What happens with script/API users?)

docs/reference.rst
```diff
@@ -891,3 +891,45 @@ Stores the number of milliseconds to wait between request retries.  By default,
...
+The server currently returns localized display names for those methods:
+
+  * Shotgun.schema_entity_read
+  * Shotgun.schema_field_read
+
+Localization is disabled by default. To enable localization, set the ``localized`` property to ``True``.
+The property cannot be set at Shotgun class instantiation. When the class is instantiated, the ``localized`` config property is set to ``False``.
```
manneohrstrom: line 905: i think we can skip this entire line! The example explains it well!

docs/reference.rst
```diff
@@ -891,3 +891,45 @@ Stores the number of milliseconds to wait between request retries.  By default,
...
+
+Example for a user whose Language preference is set to Japanese:
```
manneohrstrom: nitpick: language should have a lower case `l`

docs/reference.rst
```diff
@@ -891,3 +891,45 @@ Stores the number of milliseconds to wait between request retries.  By default,
...
+
+.. code-block:: python
+   :emphasize-lines: 8,19
```
manneohrstrom: ❤️

docs/reference.rst
```diff
@@ -891,3 +891,45 @@ Stores the number of milliseconds to wait between request retries.  By default,
...
+    },
+    ...
+    }
+    >>> sg.config.localized = True # enabling the localization
+    >>> sg.schema_field_read('Shot')
+    {
+    'sg_vendor_groups': {
+        'mandatory': {'editable': False, 'value': False},
+        # the value field (display name) is localized
+        'name': {'editable': True, 'value': '\xe3\x83\x99\xe3\x83\xb3\xe3\x83\x80\xe3\x83\xbc \xe3\x82\xb0\xe3\x83\xab\xe3\x83\xbc\xe3\x83\x97'},
```
manneohrstrom: this is a UTF-8 encoded string, right? I think we should mention that as part of the docs.

Also, it would be great if @willis102 did a quick CR on this to make sure that this approach still make sense for Python 3!

docs/reference.rst
manneohrstrom: *namnes* (typo)

manneohrstrom: suggestion: return display namnes in user's language -> return field display names in the current user's language.

docs/reference.rst
```diff
@@ -891,3 +891,45 @@ Stores the number of milliseconds to wait between request retries.  By default,
 
 In the case that both this environment variable and the config's ``rpc_attempt_interval`` property are set, the value in ``rpc_attempt_interal`` will be used.
 
+************
+Localization
+************
+
+Shotgun API offers the ability to return display namnes in user's language.
+The server currently returns localized display names for those methods:
```
manneohrstrom: I suggest changing this line (and removing the two bullet points below) to "This functionality is currently supported by the methods `schema_entity_read` and `schema_field_read`."

shotgun_api3/shotgun.py
```diff
@@ -598,6 +599,7 @@ def __init__(self,
         self.config.convert_datetimes_to_utc = convert_datetimes_to_utc
         self.config.no_ssl_validation = NO_SSL_VALIDATION
         self.config.raw_http_proxy = http_proxy
+        self.config.localized = False
```
manneohrstrom: i don't think we need this line - it's already initialized in the constructor for `_Config`, right? If we do need it, please explain the reason why with a comment!

shotgun_api3/shotgun.py
```diff
@@ -1818,6 +1820,9 @@ def schema_entity_read(self, project_entity=None):
             ``{'type': 'Project', 'id': 3}``
         :returns: dict of Entity Type to dict containing the display name.
         :rtype: dict
+
+        .. note::
+            The returned display names for this method will be localized when the ``localize`` Shotgun config property is set to ``True``. See :ref:`localization` for more informations.
```
manneohrstrom: great!

manneohrstrom: nitpick: typo: informations -> information.

shotgun_api3/shotgun.py
```diff
@@ -1917,6 +1922,9 @@ def schema_field_read(self, entity_type, field_name=None, project_entity=None):
         .. note::
             If you don't specify a ``project_entity``, everything is reported as visible.
 
+        .. note::
+            The returned display names for this method will be localized when the ``localize`` Shotgun config property is set to ``True``. See :ref:`localization` for more informations.
```
manneohrstrom: nitpick: typo: informations -> information.

tests/test_client.py
```diff
@@ -184,6 +184,28 @@ def test_authorization(self):
...
+        """Localization header passed to server when setting SG localization config to True"""
+        self.sg.config.localized = True
+
+        self.sg.info()
+
+        args, _ = self.sg._http_request.call_args
+        (_, _, _, headers) = args
+        expected_header_value = "auto"
+
+        self.assertEqual("auto", headers.get("locale"))
```
manneohrstrom: I think we probably should test whether the schema content coming back from the server is actually localised. Is this easy to do?

docs/reference.rst
palisir: Thanks for all the suggestions ! I updated the documentation.

tests/test_client.py
palisir: It's not hard to do (not sure if we can set the site language from the API, though) but then it would test the server, not the client.
IMO if the requests is made with the right header, the job is done.
Also, the localization on the server is already tested with Cypress on top of Ruby unit tests.

shotgun_api3/shotgun.py
palisir: That's correct, I removed the line

tests/test_client.py
manneohrstrom: that makes sense! I was thinking about testing the "receiving end" of the transport, eg. asserting that the received object is a `str` and that it's utf-8 encoded etc. Maybe this is one for @willis102 to weigh in on - do you think this test would be meaningful in a py3 cpntext?

PR #209, opened 2019-09-09
SG-10151: set field visibility via API

jfboismenu:
Adds support for updating field visibility through the API.

This can be achieved by the following code:

```python
sg.schema_field_update("Asset", "sg_my_custom_field", {"visible": False}, project)
```

where `project` is a dictionary referring to project in which we want to hide "sg_my_custom_field". Restoring visibility can be done via

```python
sg.schema_field_update("Asset", "sg_my_custom_field", {"visible": True}, project)
```

Note: The build is currently failing because the feature hasn't been released in Shotgun. Once the feature is released the test will be updated to the right version check and the feature commited.

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -1996,7 +1996,7 @@ class Shotgun(object):
 
         return self._call_rpc("schema_field_create", params)
 
-    def schema_field_update(self, entity_type, field_name, properties):
+    def schema_field_update(self, entity_type, field_name, properties, project_entity=None):
         """
         Update the properties for the specified field on an entity.
 
@@ -2014,6 +2014,12 @@ class Shotgun(object):
         :param field_name: Internal Shotgun name of the field to update.
         :param properties: Dictionary with key/value pairs where the key is the property to be
             updated and the value is the new value.
+        :param dict project_entity: Optional Project entity specifying which project to modify the
+            ``visible`` property for. If the ``visible`` is present in ``properties`` and
+            ``project_entity`` is not set, an exception will be raised. Example:
+            ``{'type': 'Project', 'id': 3}``
+        :param project_entity: Link to the project we wish to change the field visibility for.
+            This is necessary only when updating the ``visible`` property of a field.
         :returns: ``True`` if the field was updated.
         :rtype: bool
         """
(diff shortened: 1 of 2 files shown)


shotgun_api3/shotgun.py
```diff
@@ -2026,7 +2032,7 @@ def schema_field_update(self, entity_type, field_name, properties):
                 for k, v in six.iteritems((properties or {}))
             ]
         }
-
+        params = self._add_project_param(params, project_entity)
```
jfboismenu: The server side simply expects the project to be passed in.

shotgun_api3/shotgun.py
```diff
@@ -2014,6 +2014,12 @@ def schema_field_update(self, entity_type, field_name, properties):
         :param field_name: Internal Shotgun name of the field to update.
         :param properties: Dictionary with key/value pairs where the key is the property to be
             updated and the value is the new value.
+        :param dict project_entity: Optional Project entity specifying which project to modify the
+            ``visible`` property for. If the ``visible`` is present in ``properties`` and
+            ``project_entity`` is not set, an exception will be raised. Example:
+            ``{'type': 'Project', 'id': 3}``
+        :param project_entity: Link to the project we wish to change the field visibility for.
```
willis102: Looks like you've documented `project_entity` twice here.

willis102: Also, does `project_entity` only apply for `visible` property?  Is that something we want to call out more explicitly?

tests/test_api.py
```diff
@@ -2678,6 +2678,73 @@ def test_multiple_latest_filters(self):
                           self.sg.find,
                           "Version", filters, fields=fields, additional_filter_presets=additional_filters)
 
+    def test_modify_visibility(self):
+        """
+        Ensure the visibility of a field can be edited via the API.
+        """
+        # If the version of Shotgun is too old, do not run this test.
+        # TODO: Update this with the real version number once the feature is released.
```
willis102: Just highlighting this TODO.

tests/test_api.py
```diff
@@ -2678,6 +2678,73 @@ def test_multiple_latest_filters(self):
...
+        )
+
+        # Restore the visibility on the field.
+        self.sg.schema_field_update("Asset", field_name, {"visible": True}, project_1)
+        self.assertEqual(
+            {"value": True, "editable": True},
+            self.sg.schema_field_read("Asset", field_name, project_1)[field_name]["visible"]
+        )
+
+
```
willis102: nitpick: too many newlines?

shotgun_api3/shotgun.py
jfboismenu: Good point. Let me know what you think of my update.

PR #215, opened 2019-10-23
SG-13267: Config fixes for Mockgun

jfboismenu:
Mockgun never initialized the values inside the `_Config` instance, which is causing some issues in Python 3 because of `None` values being passed into `urllib` methods.

I've added the required tests.


willis102:
I'm going to merge this onto [SG-13264-update_sgsix](https://github.com/shotgunsoftware/python-api/tree/SG-13264-update_sgsix) and I'll be updating tk-core with the latest from there.

PR #217, opened 2019-11-04
SG-37203 [Mockgun] Ensure string comparison are case insensitive

rlessardrodeofx:
Hi!

We recently realized that for text field comparisons, Mockgun was case sensitive where a real Shotgun instance was always case insensitive. 

This seemed standard behavior so I updated Mockgun to match.
I've also added appropriate tests and new ones to increase text field comparison test coverage.

Let me know what you think.


kporangehat:
It's almost 4+ years. This feels pretty foundational to bringing mockgun functionality closer to "real world" use cases. Any update on getting this merged?

kporangehat:
🙏

PR #226, opened 2020-05-08
Add timeout support on upload operations

greenwolffr:
**Adds the support for timeout on upload request.**

According to this old post:
https://groups.google.com/a/shotgunsoftware.com/d/msg/shotgun-dev/P-YXTWs9mp0/tzXtkIALt1sJ

It was not added to shotgun API due to a backward compatibility necessity with python 2.4.
Since python 2.6 is now a requirement for the API to work it does not create an issue anymore and can be added.

It has became a priority due to the COVID situation, Shotgun creates a lot of timeouts when uploading thumbnails and playblasts.

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -3976,7 +3976,10 @@ class Shotgun(object):
             request.add_header("Content-Type", content_type)
             request.add_header("Content-Length", size)
             request.get_method = lambda: "PUT"
-            result = opener.open(request)
+            if self.config.timeout_secs is not None:
+                result = opener.open(request, timeout=self.config.timeout_secs)
+            else:
+                result = opener.open(request)
             etag = result.info()["Etag"]
         except urllib.error.HTTPError as e:
             if e.code == 500:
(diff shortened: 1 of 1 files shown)


greenwolffr:
I am waiting for some test, but I will also add a SSLError except to raise a ShotgunError instead of a Fault.

greenwolffr:
Hi @jfboismenu , we are experiencing timeouts when uploading medium playblast video files size (~3MB). 

This happens every day like 10-20 times a day according to our wranglers, mostly during business time when there is a lot of traffics. We already confirmed that it's due to an HTTP timeout issue, but since upload does not have time out configured, the request never ends and scripts hangs forever. 

This patch fixed the issue on our side, with this modification we can now detect timeouts and retry the request.

jfboismenu:
Hi!
Alright, I'll discuss this with the team, this sounds totally fair!

greenwolffr:
Hi, thanks for considering the request.

Yes for now things seems to be in control on our side. The only pitfall we had for now, is that having such a patch make difficult for us to update safely the API.

PR #235, opened 2020-11-03
SG-19308 bundle certifi

pscadding:
This PR bundles the Python module certifi with the API. The reason for doing this is so that the API ships with the latest certs at the time, and uses them by default. This should hopefully remove the issue where clients are using old Python interpreters that are bundled with out of date certs.

I haven't updated/created a software credits and requirements file for this change, as we are going to be adding one anyway as part of a separate ticket, so I felt it was better to leave that for now.


jfboismenu (approved):
Great work!
image

PR #236, opened 2020-11-12
SG-19789 licensing

pscadding:
This pr adds the licence information to a new software credits file.

I also found that the `xmlrpclib` packaged module did not appear to be used anymore. We were importing it through six instead. I've removed the module and the tests all appear fine. @jfboismenu do you have any concerns about this or know of anything I might have not thought of?


jfboismenu (approved):
Looks good to me. Have you had a change to take a look at the failing test? Any chance we can fix it quickly?

jfboismenu (changes requested):
I think we're missing a requirements.txt file here actually.

pscadding:
The requirements.txt was already added when I added certifi, it is in `shotgun_api3/lib/requirements.txt`

jfboismenu (approved):

eshokrgozar (approved):

PR #238, opened 2020-11-17
Ensure all files are py2/py3 syntactically compatible

zsimic:
This should allow to get closer to a place where `pip install git+https:...` should work with shotgun-api


shotgun_api3/lib/mimetypes.py
```diff
@@ -568,14 +568,14 @@ def _default_mime_types():
 """
 
     def usage(code, msg=''):
-        print USAGE
-        if msg: print msg
+        print(USAGE)
+        if msg: print(msg)
```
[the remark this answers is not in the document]
zsimic: Totally agree with hound bot, I left it as close to original as possible in this iteration however on purpose.
I'll gladly straighten this out though

shotgun_api3/lib/httplib2/python3/socks.py
```diff
@@ -434,7 +434,7 @@ def __negotiatehttp(self, destaddr, destport):
         wrote_host_header = False
         wrote_auth_header = False
         if self.__proxy[6] != None:
-            for key, val in self.__proxy[6].iteritems():
+            for key, val in self.__proxy[6].items():
```
zsimic, line 437: Similar change can be done in `../python2/socks.py`, intentionally left that one as-is (as it's still syntactically OK, and presumably this one is the one selected at runtime for py3)

shotgun_api3/lib/mimetypes.py
```diff
@@ -568,14 +568,14 @@ def _default_mime_types():
...
         sys.exit(code)
 
     try:
         opts, args = getopt.getopt(sys.argv[1:], 'hle',
                                    ['help', 'lenient', 'extension'])
-    except getopt.error, msg:
+    except getopt.error as msg:
         usage(1, msg)
```
zsimic, line 579: https://python-future.org/compatible_idioms.html#catching-exceptions

shotgun_api3/lib/mimetypes.py
```diff
@@ -590,9 +590,9 @@ def usage(code, msg=''):
     for gtype in args:
         if extension:
             guess = guess_extension(gtype, strict)
-            if not guess: print "I don't know anything about type", gtype
-            else: print guess
+            if not guess: print("I don't know anything about type %s" % gtype)
```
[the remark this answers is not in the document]
zsimic: Using a simple `print()` is compatible with py2/py3 (without the need for `from __future__ import print_function`).

The `__future__` import is needed if/when a more advanced usage of print function is done, like for example `print("foo", end=" ")`

shotgun_api3/lib/xmlrpclib.py
```diff
@@ -188,8 +189,8 @@ def _stringify(string):
 #__version__ = "1.0.1"
 
 # xmlrpc integer limits
-MAXINT =  2L**31-1
-MININT = -2L**31
+MAXINT =  2**31-1
+MININT = -2**31
```
zsimic, line 193: No impact: `2L` is strictly equivalent to `2` in py2 (it's a small int, the `L` was never needed)

shotgun_api3/lib/xmlrpclib.py
```diff
@@ -289,40 +290,60 @@ def __repr__(self):
...
 if _bool_is_builtin:
     boolean = Boolean = bool
-    # to avoid breaking code which references xmlrpclib.{True,False}
-    True, False = True, False
+    try:
+        # TODO: remove this when py2 support is completely dropped
+        # to avoid breaking code which references xmlrpclib.{True,False}
+        True, False = True, False
+
+    except SyntaxError:
```
zsimic, line 298: Not sure what this `True/False` trickery is about, in py3 assignment to keyword is not allowed any more.

This should have the code behave like it used in py2, and won't fail to parse with py3, however you may need to re-examine what to do in py3 in this case.

Ie: this code is py2 equivalent, but no guarantees it'll do what you want in py3

shotgun_api3/lib/xmlrpclib.py
```diff
@@ -289,40 +290,60 @@ def __repr__(self):
...
-            return cmp(self.value, other)
-        
+
+            if self.value is None:
+                return 0 if other is None else -1
+
+            if other is None:
+                return 1
+
+            return self.value.__cmp__(other)
```
zsimic, line 325: `cmp()` is gone, this code snippet makes things work exactly like before in py2.

In py3 however -> I think defining `__cmp__()` is a no-op

shotgun_api3/lib/xmlrpclib.py
```diff
@@ -142,14 +142,15 @@
 from types import *
 import socket
 import errno
-import httplib
```
zsimic, line 145: Removed this because it seemed like an unused import (no mention of `httplib` in this file).
If it is needed due to external usage, it can be changed to something like:
```
try:
    import httplib

except ImportError:
    httplib = None  # What to do in py3?
```

carlos-villavicencio-adsk:
We're [dropping support for Python 2](https://community.shotgridsoftware.com/t/important-notice-for-end-of-april-2024-reminder-breaking-changes-for-python-2-7/18079). We appreciate your contribution.

PR #255, opened 2021-11-29
SG-21170-ROUND3 [ZD to AKN] Update links referenced from code to the old support site to point to AKN

000paradox000:
This PR replaces ZD urls to AKN and developer urls


mathurf (approved):
LGTM!
:shipit:

PR #259, opened 2022-01-16
SG-25029 Add Python 3.9 coverage to Python API Azure Pipelines tests

000paradox000:
This PR add support for python 3.9 to the python-api. As an extra this PR removes some warnings when testing.

The changes are:

- Call the proper method for encoding according to python version: encodestring and encodebytes
- Call the proper method for spliturl according to python verson
- Fix configparser when running with scaped characters in python 2.7
- Replace self.assertEquals with self.assertEqual when unittesting


mathurf (approved):
LGTM! Nice job!

I also tested with some little Python API script with python 2.7 and 3.9 and everything looks good.

:shipit:

PR #265, opened 2022-05-31
mimetypes.py: fix incompatibility with python3 due to old 'print' keyword and exception 'comma' syntax

uikokahonia:
Changing print statements from the old py2 'print' keyword syntax into the newer 'print' function syntax, also changing one try-except exception catching from the old 'comma' syntax to the newer 'as' keyword syntax. Now this module can be imported in both python3 and python2 safely, prior to that it was not compatible with python3.

Note: the 'as' assignment within an exception catching is not compatible with python <= 2.5 but it is better to have this module compatible with python3 than with python <= 2.5 (https://peps.python.org/pep-3110/)

### DIFF
--- a/shotgun_api3/lib/mimetypes.py
+++ b/shotgun_api3/lib/mimetypes.py
@@ -568,14 +568,14 @@ More than one type argument may be given.
 """
 
     def usage(code, msg=''):
-        print USAGE
-        if msg: print msg
+        print(USAGE)
+        if msg: print(msg)
         sys.exit(code)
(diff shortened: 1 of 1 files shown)


carlos-villavicencio-adsk:
We're [dropping support for Python 2](https://community.shotgridsoftware.com/t/important-notice-for-end-of-april-2024-reminder-breaking-changes-for-python-2-7/18079). This module will be removed in the future. We appreciate your contribution.

PR #270, opened 2022-11-17
SG-29004 Upgrade deprecated azure pipelines machines

000paradox000:
- Replace MacOS 10.11 with 12
- Replace WIndows 2019 with 2022

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -42,7 +42,7 @@ variables:
 trigger:
   branches:
     include:
-    - SG-29004_validate_or_upgrade_if_necessary_deprecated_azure_pipelines_machines
+    - master
 pr:
   branches:
     include:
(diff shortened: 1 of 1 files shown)


mathurf (commented):
Why macOS-11, why not macOS-12 ?

mathurf (approved):
LGTM!

julien-lang:
We don't want to switch to Ubuntu 22 in this one as well?
No that this is the topic here but since you change the Windows image already and you aligned all images to the latest OS version in https://github.com/shotgunsoftware/tk-ci-tools/pull/32, it would make sense to do it here too.

000paradox000:
> We don't want to switch to Ubuntu 22 in this one as well? No that this is the topic here but since you change the Windows image already and you aligned all images to the latest OS version in [shotgunsoftware/tk-ci-tools#32](https://github.com/shotgunsoftware/tk-ci-tools/pull/32), it would make sense to do it here too.

Hi Julien thanks for the comment. In python-api we had to force ubuntu-20.04 for python-2.7 but the others images went well with latest versions. This will be solved once we deprecate python 2.

julien-lang:
> In python-api we had to force ubuntu-20.04 for python-2.7

Oh I see. Thanks for the information.

PR #276, opened 2022-12-14
SG-21915 Mockgun returns dictionary list instead of individual item

000paradox000:
This is a draft PR testing the effects of returning a dict instead of a list of dict in update method of mockgun (python-api mock class).


gplsteph:
hello, quick heads up that you will break unit tests from customers which had to deal with this if you move forward with these changes. Unfortunate, but since it has been publicly available for years, we had to deal with it on our side ;)

PR #285, opened 2023-04-24
SG-30702 Skip SG-MIM entities

carlos-villavicencio-adsk:


julien-lang (approved):
LGTM!

I don't really see why this git ignore but again, I am not a git ignore fan... ;)

carlos-villavicencio-adsk:
@julien-lang for the record. Running tests locally generated this file which should not be part of the git tree.

PR #292, opened 2023-08-04
SG-31392 Add doc for PublishedFiles preset filters

dsauve-adsk:


docs/reference.rst
```diff
@@ -570,6 +570,23 @@ with which parameters.
...
+                                                        set of Versions, this preset allows
+                                                        to find the latest PublishedFile created
+                                                        for each of these Versions.
+
+                                                      'VERSION_NUMBER':
+                                                        When dealing with multiple
+                                                        PublishedFiles associated to a group
+                                                        of entities, returns only the
+                                                        PublishedFile  with the highest
+                                                        version_number.
```
shaynacohen, line 589:
```suggestion
    Published   LATEST            [string] latest_by  'ENTITIES_CREATED_AT':
    Files                                               When dealing with multiple
                                                        PublishedFiles associated to a group
                                                        of entities, this will return only the
                                                        last PublishedFiles created for each
                                                        entity.
                                                        For example, when dealing with a
                                                        set of Versions, this preset allows you
                                                        to find the latest PublishedFile created
                                                        for each of these Versions.

                                                      'VERSION_NUMBER':
                                                        When dealing with multiple
                                                        PublishedFiles associated with a group
                                                        of entities, this returns only the
                                                        PublishedFile with the highest
                                                        version_number.
```

shaynacohen: @dsauve-adsk something strange happened when I added my suggestion, but I edited the text slightly. I would not commit this suggestion since it looks like the formatting isn't right. But, can you add the updated text?

shaynacohen (changes requested):

docs/reference.rst
dsauve-adsk: Just did the recommendations, (I also applied it where I copied the text) it changed the formatting so it is hard to validated, but everything should be there...

shaynacohen (approved):
Looks great!

PR #295, opened 2023-08-15
SG-31340 Deprecation of Python 2

minna-feng:

PR #297, opened 2023-08-21
SG-32061 Python3.7+ Syntax

mathbou:
Hi there,

I just saw through [SG-31340-deprecation-py2](https://github.com/shotgunsoftware/python-api/pull/295) that py2 has officially been deprecated, so I updated the API to embrace a py3.7+ syntax

The biggest changes:
- six and sgsix aren't needed anymore
- No need for two versions of httplib2 next to each other, we use the PyPI version instead. This also means pyparsing can be removed from the lib.
- certifi are also fetched from PyPI
- replace printf syntax with fstrings and `str.format`


mathbou:
> Thank you for that! Do you know why unit tests are failing? It seems they share the same root cause.

I think I found the reason: CI tests relies on [variables](https://github.com/shotgunsoftware/python-api/blob/master/azure-pipelines-templates/run-tests.yml#L93-L98) given by the `run_tests.yml` file. If tests are triggered from a PR, those value should be null, but the syntax that has been used prevent that, and instead the name of those variables were returned.

carlos-villavicencio-adsk:
Amazing! 

I have one question, since [tk-core bundles python-api module](https://github.com/shotgunsoftware/tk-core/tree/master/python/tank_vendor/shotgun_api3), would it be an issue to have dependencies such as certifi not bundled as before?

mathbou:
> I have one question, since [tk-core bundles python-api module](https://github.com/shotgunsoftware/tk-core/tree/master/python/tank_vendor/shotgun_api3), would it be an issue to have dependencies such as certifi not bundled as before?

Indeed yes, It will not work as is. I made a PR for the `tk-core` that would support this  "vendor-less" version https://github.com/shotgunsoftware/tk-core/pull/914

carlos-villavicencio-adsk:
Thank you for the clarification. If it's possible, can you split this work into two different PRs? We want to address these improvements separately. So ideally the extraction of the bundled dependencies can be reviewed outside of this scope.

mathbou:
@carlos-villavicencio-adsk 

I split this PR in two 
- The syntax is [here](https://github.com/shotgunsoftware/python-api/pull/303)
- The vendorless update [there](https://github.com/shotgunsoftware/python-api/pull/302)

A rebase will probably be required between the two merge to ensure there are no conflict between them

PR #301, opened 2023-09-08
SG-32806 Enable support for Python 3.10 in CI

carlos-villavicencio-adsk:
We need to ~~rebase #305 and~~ complete SG-32847 before testing this again.

- [x] ssl.PROTOCOL_TLS is deprecated
    - Upgrades httplib2 module
- [x] ssl.wrap_socket() is [deprecated](https://docs.python.org/3/library/ssl.html#ssl.wrap_socket), use SSLContext.wrap_socket()
    - Upgraded syntax
- [ ] module 'sre_constants' is deprecated (only on Python 3.11)
    - pyparsing should be upgraded. We can work on this later since this needs more testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
-[![VFX Platform](https://img.shields.io/badge/vfxplatform-2020-blue.svg)](http://www.vfxplatform.com/)
-[![Python 3.7](https://img.shields.io/badge/python-3.7-blue.svg)](https://www.python.org/)
+[![VFX Platform](https://img.shields.io/badge/vfxplatform-2023-blue.svg)](http://www.vfxplatform.com/)
+[![Python 3.7 3.9 3.10](https://img.shields.io/badge/python-3.7%20%7C%203.9%20%7C%203.10-blue.svg)](https://www.python.org/)
 [![Reference Documentation](http://img.shields.io/badge/doc-reference-blue.svg)](http://developer.shotgridsoftware.com/python-api)
 [![Build Status](https://dev.azure.com/shotgun-ecosystem/Python%20API/_apis/build/status/shotgunsoftware.python-api?branchName=master)](https://dev.azure.com/shotgun-ecosystem/Python%20API/_build/latest?definitionId=108&branchName=master)
 [![Coverage Status](https://coveralls.io/repos/github/shotgunsoftware/python-api/badge.svg?branch=master)](https://coveralls.io/github/shotgunsoftware/python-api?branch=master)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,5 +1,5 @@
-[![VFX Platform](https://img.shields.io/badge/vfxplatform-2020-blue.svg)](http://www.vfxplatform.com/)
-[![Python 3.7](https://img.shields.io/badge/python-3.7-blue.svg)](https://www.python.org/)
+[![VFX Platform](https://img.shields.io/badge/vfxplatform-2023-blue.svg)](http://www.vfxplatform.com/)
```
julien-lang, line 1: Just like for Python versions, we should say we support multiple VFX versions: CY2023	CY2022	CY2021	CY2020

azure-pipelines-templates/run-tests.yml
```diff
@@ -43,11 +43,19 @@ jobs:
...
       Python39:
         python.version: '3.9'
+      Python310:
+        python.version: '3.10'
+      # Note for Python 3.11. This will raise hundres of warnings on a third party module
+      # pytest_nunit/nunit.py
+      # DeprecationWarning: Use setlocale(), getencoding() and getlocale() instead
+      # This is the current behavior on the latest 1.0.3 version of this module.
+
+    maxParallel: 4
```
julien-lang, line 58: I don't know if it matters that much. But should not that be 3?

shotgun_api3/shotgun.py
```diff
@@ -4244,11 +4244,19 @@ def connect(self):
         "Connect to a host on a given (SSL) port."
         http_client.HTTPConnection.connect(self)
         # Now that the regular HTTP socket has been created, wrap it with our SSL certs.
-        self.sock = ssl.wrap_socket(
-            self.sock,
-            ca_certs=self.__ca_certs,
-            cert_reqs=ssl.CERT_REQUIRED
-        )
+        if six.PY38:
```
julien-lang, line 4247: Does that mean ">=3.8"?

carlos-villavicencio-adsk: It does

NorberMV (approved):
LGTM !

julien-lang (commented):
@carlos-villavicencio-adsk what about the Chorus warnings? Anything we can do to avoid them? Any comment pattern to silent them?

carlos-villavicencio-adsk:
> @carlos-villavicencio-adsk what about the Chorus warnings? Anything we can do to avoid them? Any comment pattern to silent them?

We cannot avoid them without completely removing the Python 2 support. The warning is on the existing legacy code we had. We can skip them, it's on the same invocation.

julien-lang (approved):

PR #305, opened 2023-09-18
SG-32806 Upgrade httplib2 to 0.22.0. Also pyparsing==3.1.1

carlos-villavicencio-adsk:
In preparation for Python 3.10 we need to upgrade these two dependencies to remove warnings on CI and potential deprecation.


julien-lang (dismissed):
LGTM!

Do you have any idea why we use httplib2 anyway?

PR #310, opened 2023-10-09
SG-31927 Flaky Test In python-api Test Work schedule

NorberMV:
Removing a flaky assertion that compares two calls to work_schedule_read() with the same parameters, as it serves no clear purpose and occasionally fails in the CI. Running this same test by making real calls to the python-api correctly returns the `WorkDayRules` defined between the given `start_date `and `end_date`.


carlos-villavicencio-adsk (approved):
Good catch! :shipit:

PR #311, opened 2023-10-11
Travis badge image is no working anymore

julien-lang:
We don't even use Travis anyway

image


carlos-villavicencio-adsk (approved):
:shipit:

PR #313, opened 2023-10-25
Fix flaky upload tests.

NorberMV:
This PR fixes flaky tests that contain the following errors:
```
- AssertionError: 'image/png' != 'image/jpeg'
- AssertionError: '304' != '200'
```
Allows upload requests with status '304' in the thumbnail upload tests, because sometimes when the response hasn't been modified, this status informs the client that the response has not changed, so the client can continue using the same cached version.

Also supports transient placeholder PNG file content types in the response payload, since sometimes it takes longer for the previously uploaded image requested to be available from media storage, and we have previously tested that the image was successfully uploaded.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -336,8 +336,8 @@ class TestShotgunApi(base.LiveTestBase):
 
         h = Http(".cache")
         thumb_resp, content = h.request(new_version.get('image'), "GET")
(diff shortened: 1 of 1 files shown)


carlos-villavicencio-adsk (approved):
LGTM :shipit:

PR #316, opened 2023-12-15
SG-33891 Retries also on 504

carlos-villavicencio-adsk:


shotgun_api3/shotgun.py
```diff
@@ -3443,8 +3443,8 @@ def _call_rpc(self, method, params, include_auth_params=True, first=False):
                 # We've seen some rare instances of SG returning 502 for issues that
                 # appear to be caused by something internal to SG. We're going to
                 # allow for limited retries for those specifically.
-                if attempt != max_attempts and e.errcode == 502:
-                    LOG.debug("Got a 502 response. Waiting and retrying...")
+                if attempt != max_attempts and e.errcode in [502, 504]:
```
julien-lang, line 3446: You might want to use HTTPStatus constants instead: `http.HTTPStatus.SERVICE_UNAVAILABLE` and `http.HTTPStatus.GATEWAY_TIMEOUT`

Not a blocker but it's usually better to use constants.

tests/test_client.py
```diff
@@ -430,12 +430,28 @@ def test_call_rpc(self):
...
+        # Test that we raise on a 5xx. This is ensuring the retries behavior
+        # in place specific to 5xx responses still eventually ends up raising.
+        # 502
+        max_attempts = 4
         d = {"results": ["foo", "bar"]}
         a = {"some": "args"}
         self._mock_http(d, status=(502, "bad gateway"))
         self.assertRaises(api.ProtocolError, self.sg._call_rpc, "list", a)
+        self.assertTrue(
+            max_attempts == self.sg._http_request.call_count,
```
julien-lang, line 442: Why not using assertEqual instead and also save a variable?
```python
 self.assertEqual(
    4,
    self.sg._http_request.call_count,
    "Call is repeated up to 3 times",
)
```

julien-lang (dismissed):

julien-lang (approved):
LGTM!

NorberMV (approved):
LGTM! 🚢

PR #318, opened 2024-01-04
SG-34199 #comment typing annotation breaks Python 2 compatibility

slackba:
We pull `shotgun_api3` directly for deployment here at TTF. We have some old shows still using Maya2019 (and thereby Python 2). This addition of hinting annotations breaks import of your packages. If it's intentional, and you are dropping Py2 support, then disregard this PR. In that case, we'll have to lock our release for our Py2 deployments (while we have them).

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -3273,7 +3273,8 @@ class Shotgun(object):
 
         return self._call_rpc("user_subscriptions_read", None)
 
-    def user_subscriptions_create(self, users: list):
+    def user_subscriptions_create(self, users):
+        # type: (list[dict[str, Union[str, list[str], None]) -> bool
         """
         Assign subscriptions to users.
 


julien-lang:
Hi @slackba, Thank you for reporting this issue!

This regression was not intentional so we are going to merge this change. For CI reasons, I had to create a clone PR: #320.

However, we are going to remove the compatibility with Python 2 in a few months. For more information about this, please read this announcement: https://community.shotgridsoftware.com/t/important-notice-for-end-of-april-2024-reminder-end-of-support-for-python-2-7/.

Also, regarding Maya, we wanted to let you know that ShotGrid Toolkit/Python API no longer supports Maya 2019. The oldest supported version is Maya 2020. For more information about Engine Supported Versions, please refer to the following documentation: https://help.autodesk.com/view/SGDEV/ENU/?guid=SGD_si_integrations_engine_supported_versions_html.

PR #319, opened 2024-01-06
SG-34359 Add field type `entity_type` to mockgun as simple str

slingshotvfx:
For testing purposes, it can be helpful to support `entity_type` field types, even if there's no validation that it's an existing Entity from the schema.

### DIFF
--- a/shotgun_api3/lib/mockgun/mockgun.py
+++ b/shotgun_api3/lib/mockgun/mockgun.py
@@ -495,6 +495,7 @@ class Shotgun(object):
                                    "checkbox": bool,
                                    "percent": int,
                                    "text": six.string_types,
+                                   "entity_type": six.string_types,
                                    "serializable": dict,
                                    "date": datetime.date,
                                    "date_time": datetime.datetime,


carlos-villavicencio-adsk:
Hello, thank you for the contribution. Can you provide a use case when this is helpful? Cheers

slingshotvfx:
Hey Carlos,

I was adding tests for code that creates ActionMenuItems on a site through the Python API. So I wanted to assert they were created in Mockgun.

The following code: 
```python
    sg_amis = mockgun.find(
        "ActionMenuItem",
        filters=[],
        fields=[
            "title",
            "entity_type",
        ],
    )
```

results in `ShotgridError: Field ActionMenuItem.entity_type: Handling for ShotGrid type entity_type is not implemented`

As far as I can tell, entity_type fields in Shotgrid are just text fields, maybe limited to a specific set of values matching the entities available in the Shotgrid site.  I suppose in Mockgun you could add some validation to verify that the value of entity_type fields exists as an entity name in the Schema, but for my purposes just setting it to a string worked great.


Thanks!

carlos-villavicencio-adsk:
I think it's a good idea to add `entity_type`. Thanks. Our team will review this and merge as soon as they can.

PR #328, opened 2024-02-06
add set_session_uuid to mockgun

slingshotvfx:
Adds set_session_uuid to mockgun.  This has no real effect, but allows code to be tested that calls the [real function](https://github.com/shotgunsoftware/python-api/blob/1666c39956207a9cde377b319c9df45bd9745eb2/shotgun_api3/shotgun.py#L2191).


barbara-darkshot:
Hi @slingshotvfx!
I'm gonna close this PR as we released a new version of [python-api](https://github.com/shotgunsoftware/python-api/releases/tag/v3.5.1) yesterday where we did some little improvements to Mockgun that should also cover this use case.
Thanks!

PR #330, opened 2024-02-09
add multi_entity_update_modes support to mockgun

slingshotvfx:
Adds `multi_entity_update_modes` support to mockgun `update()` and `batch()` methods.

per docs: https://developers.shotgridsoftware.com/python-api/reference.html#shotgun_api3.shotgun.Shotgun.update

adds support for `add`, `set`, and `remove` multi_entity_update_modes, with the default being `set`

Also adds an accompanying test to test_mockgun.py


carlos-villavicencio-adsk (approved):
LGTM. In #341 we could confirm CI tests are ✅

PR #337, opened 2024-04-16
SG-34551 validation for python3, so display_name now is only decoded …

eduardoChaucaGallegos:
Now prefix/suffix if display_name property exists is not added


shotgun_api3/shotgun.py
```diff
@@ -2587,7 +2587,10 @@ def _upload_to_sg(self, entity_type, entity_id, path, field_name, display_name,
             url = urllib.parse.urlunparse((self.config.scheme, self.config.server,
                                            "/upload/upload_file", None, None, None))
             if display_name is None:
-                display_name = os.path.basename(path)
+                if six.PY3:
```
julien-lang, line 2590: Minor comment: I'd prefer having the opposite test: if Py2:... else.
So the specific is about Py2.

julien-lang (commented):
Looks good. Are there any other fields that could have the same problem? 

Also is it possible to add unit tests?

eduardochauca90:
@julien-lang  By the moment I couldn't see other fields with that potential problem in the scope of "_upload_to_sg" function because it is the only one field that previously was encoded to utf-8(encode("utf-8")) to assurance the behavior of non-ascii unicode and utf-8 string paths during upload.
Also I believe that is not necessary a new test for that new condition because that function and complete upload flow already has a unit tests. There exists a possibility to use another python2 "runner" in settings of CI/CD, but If I'm not wrong, we only will support python3 in the next weeks

tests/test_api.py
```diff
@@ -310,6 +311,54 @@ def test_upload_download(self):
...
+    @patch('shotgun_api3.Shotgun._send_form')
+    def test_upload_to_sg(self, mock_send_form):
+        """
+        Upload an attachment tests for _upload_to_sg()
+        """
+        mock_send_form.method.assert_called_once()
+        mock_send_form.return_value = "1\n:123\nasd"
+        if 'localhost' in self.server_url:
+            print("upload / down tests skipped for localhost")
+            return
```
carlos-villavicencio-adsk, line 323: Another option could be to use pytest's [skipif](https://docs.pytest.org/en/latest/how-to/skipping.html#id1) marker.

Not a blocker, just to let you know you have this option.

tests/test_api.py
```diff
@@ -24,6 +24,7 @@
 import uuid
 import unittest
 from shotgun_api3.lib.six.moves import range, urllib
+import pytest
```
julien-lang, line 27: I don't see how this import is required.

tests/test_api.py
julien-lang: Agreed. Either a decorator or just call [skipIf](https://docs.python.org/3/library/unittest.html#unittest.skipIf):

```suggestion
        self.skipIf(
            'localhost' in self.server_url,
            "upload / down tests skipped for localhost",
        )
```

carlos-villavicencio-adsk (approved):
LGTM!

shotgun_api3/shotgun.py
```diff
@@ -2587,7 +2587,10 @@ def _upload_to_sg(self, entity_type, entity_id, path, field_name, display_name,
             url = urllib.parse.urlunparse((self.config.scheme, self.config.server,
                                            "/upload/upload_file", None, None, None))
             if display_name is None:
-                display_name = os.path.basename(path)
+                if six.PY2:
+                    display_name = os.path.basename(path)
+                else:
+                    display_name = os.path.basename(path.decode("utf-8"))
```
julien-lang, line 2593: Can you 15 lines above. There is a big comment about unicode path and we encode path in utf-8... Looks like the opposite of what's done here. We might want to refactor/clean/review all that

carlos-villavicencio-adsk (approved):

julien-lang (approved):
LGTM!

PR #342, opened 2024-05-22
SG-35165 Retry on URLError

carlos-villavicencio-adsk:
- Add retry only when encountering an URLError
- Adopt the existing strategy
- Add tests

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -500,7 +500,7 @@ class Shotgun(object):
         r"(\D?([01]\d|2[0-3])\D?([0-5]\d)\D?([0-5]\d)?\D?(\d{3})?)?$")
 
     _MULTIPART_UPLOAD_CHUNK_SIZE = 20000000
-    MAX_ATTEMPTS = 4    # Three retries on failure
+    MAX_ATTEMPTS = 3    # Retries on failure
     BACKOFF = 0.75      # Seconds to wait before retry, times the attempt number
 
     def __init__(self,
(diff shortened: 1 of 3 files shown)


shotgun_api3/shotgun.py
```diff
@@ -4250,19 +4252,28 @@ def _send_form(self, url, params):
...
+        attempt = 1
+        while attempt <= self.MAX_ATTEMPTS:
+            # Perform the request
+            try:
+                resp = opener.open(url, params)
+                result = resp.read()
+                # response headers are in str(resp.info()).splitlines()
+            except urllib.error.URLError as e:
+                LOG.debug("Got a %s response. Waiting and retrying..." % e)
+                time.sleep(float(attempt) * self.BACKOFF)
```
julien-lang, line 4264: The problem with this loop is that, in the worst case scenario, where every attempt failed, you end up sleeping before leaving the loop.

tests/test_client.py
```diff
@@ -449,7 +449,7 @@ def test_call_rpc(self):
         self._mock_http(d, status=(504, "gateway timeout"))
         self.assertRaises(api.ProtocolError, self.sg._call_rpc, "list", a)
         self.assertEqual(
-            4,
+            3,
```
julien-lang, line 452: what's this change?

carlos-villavicencio-adsk: I standardize the number of retries to be reused on the library.

julien-lang: Can you use `self.sgMAX_ATTEMPTS` instead or does it make sense to keep these 3 everywhere?

carlos-villavicencio-adsk: I think it's feasible.

PR #343, opened 2024-05-31
Minor code refactoring: make download_attachment directly use _build_opener

julien-lang:
download_attachment is the only function using set_up_auth_cookie.
Different than other requests and harder to understand.
Use global registry while the rest of the code use handler variables.


carlos-villavicencio-adsk (approved):
LGTM

eduardoChaucaGallegos (approved):

PR #345, opened 2024-06-04
SG-29997 Removing Python 2 code

eduardoChaucaGallegos:
- Removing the use of six package
 - Useful functions for shotgun.py copied to sgsix.py


shotgun_api3/lib/httplib2/__init__.py
```diff
@@ -9,24 +7,13 @@
...
-
-
-else:
-    # Generate ssl_error_classes
-    from .python2 import SSLHandshakeError as __SSLHandshakeError  # TODO: shouldn't rely on this. not public
-    ssl_error_classes = (__SSLHandshakeError,)
-    del __SSLHandshakeError
+# Generate ssl_error_classes
+import ssl as __ssl
+ssl_error_classes = (__ssl.SSLError, __ssl.CertificateError)
```
julien-lang, line 12: Do we need to keep this variable or can we just replace the code where ssl_error_classes is used by __ssl.SSLError, __ssl.CertificateError?

shotgun_api3/lib/httplib2/__init__.py
```diff
@@ -9,24 +7,13 @@
...
+del __ssl
 
-    # get the python2 fork of httplib2
-    from . import python2 as __httplib2_compat
+# get the python3 fork of httplib2
+from . import python3 as __httplib2_compat
```
julien-lang, line 16: why do we have a "python3" folder in there? Can we simplify how we bundle httplib2 in this repo?

shotgun_api3/shotgun.py
```diff
@@ -46,21 +39,18 @@
...
-    from base64 import encodestring as base64encode
+from io import BytesIO
+from http import cookiejar
+from http import client as http_client
+from xmlrpc.client import Error, ProtocolError, ResponseError
+from base64 import encodebytes as base64encode
 
+from .lib.httplib2 import Http, ProxyInfo, socks, ssl_error_classes
+from .lib.sgtimezone import SgTimezone
+from .lib import sgsix
```
julien-lang, line 53: Can/Should we get rid of that? (or rename the module to "something_utils.py"?)

shotgun_api3/shotgun.py
```diff
@@ -2136,7 +2126,7 @@ def schema_field_update(self, entity_type, field_name, properties, project_entit
             "field_name": field_name,
             "properties": [
                 {"property_name": k, "value": v}
-                for k, v in six.iteritems((properties or {}))
+                for k, v in (properties.items() if isinstance(properties, dict)  else {})
```
julien-lang, line 2129: I don't think this is correct. What you want is:
```suggestion
                for k, v in ((properties if isinstance(properties, dict) else {}).items())
```

But I think we'd better have this check done earlier in the function instead.

julien-lang (commented):
Good start.

shotgun_api3/lib/httplib2/__init__.py
julien-lang, line 1: I don't think we want to edit httplib2. This is third-party we bundle. But can we update to a newer version if we drop py2?

julien-lang (commented):
Make sure to look #202 when addressing this PR

HISTORY.rst
```diff
@@ -4,6 +4,12 @@ Flow Production Tracking Python API Changelog
 
 Here you can see the full list of changes between each Python API release.
 
+v3.7.0 (2024 Jun 10)
+===================
+- Drop support for Python 2.x.x
+- Drop use of six package
+- Documentation update
```
julien-lang, line 11: We only change this file when we prepare a release of python-api. Not when we work on the code.

setup.py
```diff
@@ -18,16 +18,12 @@
...
 script_args = sys.argv[1:]
-if (sys.version_info[0] <= 2) or (sys.version_info[0] == 2 and sys.version_info[1] <= 5):
-    if 'install' in script_args and '--no-compile' not in script_args:
-        script_args.append('--no-compile')
 
 
 setup(
     name='shotgun_api3',
-    version='3.6.0',
+    version='3.7.0',
```
julien-lang, line 26: No, not in this PR

shotgun_api3/lib/httplib2/__init__.py
julien-lang: Actually, it was manually updated by #202 so we might want to clean up (or simply update to a newer version)

shotgun_api3/shotgun.py
```diff
@@ -3575,7 +3546,7 @@ def _make_call(self, verb, path, body, headers):
             attempt += 1
             try:
                 return self._http_request(verb, path, body, req_headers)
-            except ssl_error_classes as e:
+            except (SSLError, CertificateError) as e:
                 # Test whether the exception is due to the fact that this is an older version of
                 # Python that cannot validate certificates encrypted with SHA-2. If it is, then
```
julien-lang, line 3551: This comment talks about Python 2.5 and below. We might want to clean that up. I started that in #347.

julien-lang (commented):
I am starting to get scared by this PR. This is super big and going all over the place. We might want to chunk it.

shotgun_api3/shotgun.py
```diff
@@ -4320,78 +4291,68 @@ class FormPostHandler(urllib.request.BaseHandler):
...
         for (key, value) in params:
-            if not isinstance(value, six.string_types):
+            if not isinstance(value, str):
                 # If value is not a string (e.g. int) cast to text
-                value = six.text_type(value)
-            value = six.ensure_text(value)
-            key = six.ensure_text(key)
+                value = str(value)
+            value = sgutils.ensure_text(value)
+            key = sgutils.ensure_text(key)
```
carlos-villavicencio-adsk, line 4326: same here, I think we can safely replace `str` casting method isntead.

carlos-villavicencio-adsk (commented):
Looks good so far. It would be great to depend less from six or sgsix functions.

I also opened a few questions.

shotgun_api3/lib/sgutils.py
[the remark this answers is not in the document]
carlos-villavicencio-adsk: @eduardoChaucaGallegos  Crazy idea: can we leave these methods on six.py for now and keep a mini-version of six.py only with the methods implemented in this file?

My intention is to avoid renaming the package and impact the TK components that depends on these functions.

We can discuss this internally.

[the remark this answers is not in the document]
julien-lang: Not so crazy, I kind of like that.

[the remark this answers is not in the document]
eduardoChaucaGallegos: I'm going to revert changes in this MR to only apply steps for first phase (that mini-version of six.py can be sgutils.py )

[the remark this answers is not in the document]
carlos-villavicencio-adsk: > that mini-version of six.py can be sgutils.py

Yes, but keeping the same name of six.py

That way, dependant component won't need to rename the import for ensure_*

PR #346, opened 2024-06-05
SG-34910 Fixup SSLEOFError exception

julien-lang:
Use the existing re-attempt strategy in case of SSLEOFError

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -3625,6 +3625,7 @@ class Shotgun(object):
 
                 self._close_connection()
                 if attempt == max_rpc_attempts:
+                    LOG.debug("Request failed.  Giving up after %d attempts." % attempt)
                     raise
             except Exception:
                 self._close_connection()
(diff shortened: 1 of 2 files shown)


eduardoChaucaGallegos (approved):

shotgun_api3/shotgun.py
```diff
@@ -3594,6 +3587,22 @@ def _make_call(self, verb, path, body, headers):
...
                 return self._http_request(verb, path, body, req_headers)
+            except ssl.SSLEOFError as e:
+                # SG-34910 - EOF occurred in violation of protocol (_ssl.c:2426)
+                # This issue seems to be related to proxy and keep alive.
+                # It looks like, sometimes, the proxy drops the connection on
+                # the TCP/TLS level despites the keep-alive. So we need to close
+                # the connection and make a new attempt.
+                LOG.debug("SSLEOFError: {}".format(e))
+                self._close_connection()
+                if attempt == max_rpc_attempts:
```
carlos-villavicencio-adsk, line 3598: `self.config.max_rpc_attempts` is hardcoding the value of `3`.  Maybe we can now get rid of this one and use `self.MAX_ATTEMPTS` instead. WDYT?

carlos-villavicencio-adsk (approved):
LGTM

shotgun_api3/shotgun.py
julien-lang: Hardcoded yes and no. It's the default value of the config object but users can select a different value.

Also, I see that this variable is still used at other places. So I kind of want to keep it as is for the moment.

PR #347, opened 2024-06-05
SG-30084 Custom revert of 3e72b44a to fix the SSLEOFError

julien-lang:
Could not "git revert" because the files have changed too much

All this code in 3e72b44a was created to support Python versions 2.5 and below (the Python SSL module was added in Python 2.6: https://docs.python.org/2.7/library/ssl.html)

Since we don't support these versions of Python anymore and this creates a problem with SSLEOFError, let's just revert it all.


shotgun_api3/shotgun.py
```diff
@@ -112,6 +112,8 @@ def _is_mimetypes_broken():
 have a self-signed internal certificate that isn't included in our certificate bundle, you may
 not require the added security provided by enforcing this.
 """
+if os.environ.get("SHOTGUN_DISABLE_SSL_VALIDATION", False):
+    NO_SSL_VALIDATION = True
```
julien-lang, line 116: Do we really want to bring this variable back?

PR #349, opened 2024-06-14
SG-35018 Remove Ticket entity reference and prepare this to run in CI

carlos-villavicencio-adsk:
Ticket entity is disabled by default on ShotGrid (it seems so).

image

I'm making unit tests dependant-free on this entity. It seems they were enabled when these tests were implemented, but now this is making it harder to test locally with fresh SG sites.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -1521,6 +1521,36 @@ class TestFind(base.LiveTestBase):
         result = self._id_in_result('Version', filters, self.version['id'])
         self.assertFalse(result)
 
+    def test_in_relation_comma_list(self):
+        """
+        Test that 'in' relation using commas (old format) works with list fields.
+        """
+        filters = [['frame_count', 'in', self.version['frame_count'], 33],
+                   ['project', 'is', self.project]]
+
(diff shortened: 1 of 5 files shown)


eduardoChaucaGallegos (dismissed):

tests/test_api.py
```diff
@@ -1521,36 +1521,6 @@ def test_not_in_relation_float(self):
...
         self.assertFalse(result)
 
-    def test_in_relation_comma_list(self):
-        """
-        Test that 'in' relation using commas (old format) works with list fields.
-        """
-        filters = [['sg_priority', 'in', self.ticket['sg_priority'], '1'],
-                   ['project', 'is', self.project]]
-
-        result = self._id_in_result('Ticket', filters, self.ticket['id'])
```
julien-lang, line 1531: Not equivalent with Version on these ones ?

carlos-villavicencio-adsk: Good call! I just replace them and got these tests back. Thank you!

julien-lang (approved):
LGTM!

PR #350, opened 2024-06-20
SG-35018 Condition auth for Jenkins environment

carlos-villavicencio-adsk:
### Checklist 

- [x] Use login/password for authentication when running on `SG_JENKINS`.
- [x] Unskip and fix old tests
- [x] Skip datetime tests on Jenkins (they also fail locally)
- [x] Refactor helper method to reduce sleep time in e861767

### Centralized authentication

Now all tests use the same authentication credentials. By default, it uses and expects a script name/key pair provided on the config file. However, a special flag `SG_JENKINS` is considered to use username/password only for SG sites that haven't created a script.

### Improved test time

Before | After
-- | --
image | image

### Test durations

Argument `durations` was added to pytest to detect the slowest tests and improve them in the future

image

### How to reproduce Jenkins environment locally:

Step 1: remove `script_name`/`api_key` from config file

Step 2: add `human_login` / `human_password` to config file

Step 3: run

```bash
export SG_JENKINS=1
pytest tests/test_api.py
```

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -15,5 +15,4 @@
 [run]
 source=shotgun_api3
 omit=
-    shotgun_api3/lib/httplib2/*
-    shotgun_api3/lib/six.py
+    shotgun_api3/lib/*
--- a/.coveragerc
+++ b/.coveragerc
@@ -15,4 +15,7 @@
 [run]
 source=shotgun_api3
 omit=
-    shotgun_api3/lib/*
+    shotgun_api3/lib/httplib2/*
+    shotgun_api3/lib/six.py
+    shotgun_api3/lib/certify/*
+    shotgun_api3/lib/pyparsing.py
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -2539,6 +2539,10 @@ class TestNoteThreadRead(base.LiveTestBase):
         """
         if not self.sg.server_caps.version or self.sg.server_caps.version < (6, 2, 0):
             return
+        
+        user_entity = "ApiUser"
+        if self.config.jenkins:
+            user_entity = "HumanUser"
 
         # create note
         note = self.sg.create("Note", {"content": "Test!", "project": self.project})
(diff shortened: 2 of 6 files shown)


tests/test_api.py
```diff
@@ -2169,7 +2173,7 @@ def test_human_user_sudo_auth_fails(self):
                                  login=self.config.human_login,
                                  password=self.config.human_password,
                                  http_proxy=self.config.http_proxy,
-                                 sudo_as_login="blah")
+                                 sudo_as_login="norberto.moreno@autodesk.com")
```
eduardoChaucaGallegos, line 2176: can we handle this email in the setup?

tests/test_api.py
```diff
@@ -2520,6 +2527,10 @@ def _check_attachment(self, data, attachment_id, additional_fields):
                                            [["id", "is", attachment_id]],
                                            list(expected_fields))
 
+        # remove images before comparison
+        if "this_file" in attachment_data:
+            attachment_data["this_file"].pop("url")
+            data["this_file"].pop("url")
```
eduardoChaucaGallegos, line 2533: Is it probable that "this_file" key doesn't exist in "data" dict? so we should check if "this_file" exists in "data"

tests/test_api.py
carlos-villavicencio-adsk: I'll revert this. Sorry about that.

eduardoChaucaGallegos (dismissed):

julien-lang (dismissed):

tests/test_api.py
```diff
@@ -1010,6 +1006,8 @@ def test_set_date(self):
         self.assertEqual(expected, actual)
 
     def test_set_date_time(self):
+        if self.config.jenkins:
+            self.skipTest("Skipping test on Jenkins. locked_until not updating.")
```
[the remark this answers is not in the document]
carlos-villavicencio-adsk: It's not. I want to read `self.config.jenkins`, which is unavailable in the global scope.

tests/test_api.py
```diff
@@ -1853,6 +1852,10 @@ def test_following(self):
 
 
 class TestErrors(base.TestBase):
+    def setUp(self):
+        auth_mode = "HumanUser" if self.config.jenkins else "ApiUser"
+        super(TestErrors, self).setUp(auth_mode)
```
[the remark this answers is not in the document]
carlos-villavicencio-adsk: Not really, I just didn't want to refactor existing logic for now.

.coveragerc
```diff
@@ -15,5 +15,4 @@
 [run]
 source=shotgun_api3
 omit=
-    shotgun_api3/lib/httplib2/*
-    shotgun_api3/lib/six.py
+    shotgun_api3/lib/*
```
julien-lang, line 18: I am not sure about that. There are files in this folder that belong to us:

- sgsix.py
- sgtimezone.py
- mockgun

tests/test_api.py
```diff
@@ -2532,6 +2541,10 @@ def test_simple(self):
         if not self.sg.server_caps.version or self.sg.server_caps.version < (6, 2, 0):
             return
 
+        user_entity = "ApiUser"
+        if self.config.jenkins:
+            user_entity = "HumanUser"
```
julien-lang, line 2546: Minor: you could use a ternary assignment here instead. Would make sense because you did similar above.

julien-lang (commented):
Looks good. Only a few minor comments.

.coveragerc
carlos-villavicencio-adsk: Good catch

julien-lang (approved):

PR #355, opened 2024-08-26
Fix incorrect Shotgun import in Python API example

sungbinlee:
- Corrected the import statement for the Shotgun class to use the proper module path 'shotgun_api3.Shotgun'


carlos-villavicencio-adsk (approved):
Thanks!

PR #358, opened 2024-09-05
Remove min supported py version on docs

carlos-villavicencio-adsk:


README.md
```diff
@@ -10,10 +10,6 @@ Autodesk provides a simple Python-based API for accessing Flow Production Tracki
 
 The latest version can always be found at http://github.com/shotgunsoftware/python-api
 
-## Minimum Requirements
-
-* Python v3.7
```
julien-lang, line 15: I think I'd prefer to keep this section and renaming Supported version with the full list.

I know it duplicates the badges part but that would give us a permalink (section link) we can share. Text info is easier to share than badges.
Also I know we have our general doc for supported version. But Python-API is not Toolkit. So I think it's worth duplicating the info here.

carlos-villavicencio-adsk: This makes maintenance difficult because this is a section not all components have... and we can easily forget to update it.

Maybe we can link to https://help.autodesk.com/view/SGDEV/ENU/?guid=SGD_si_platform_supported_versions_html instead?

PR #359, opened 2024-11-06
SG-36700 Upgrade Azure Pipelines vm images

carlos-villavicencio-adsk:


azure-pipelines.yml
```diff
@@ -58,14 +58,14 @@ jobs:
 - template: azure-pipelines-templates/run-tests.yml
   parameters:
     name: Linux
-    vm_image: 'ubuntu-20.04'
+    vm_image: 'ubuntu-latest'
```
eduardoChaucaGallegos, line 61: why to the latest version and not a fixed version?

carlos-villavicencio-adsk: Since this is a widely used library by the community, I'd prefer to be able to detect breaking changes early and comply with modern operating systems. Does it make sense?

julien-lang (approved):

eduardoChaucaGallegos (approved):

PR #360, opened 2024-11-15
SG-36677 Optimize payload by prevent unnecessary data

carlos-villavicencio-adsk:
## Summary
Remove unnecessary data in the payload when [combining related queries](https://developers.shotgridsoftware.com/python-api/cookbook/usage_tips.html#combining-related-queries) before sending it to the server. This would improve overall performance decreasing network latency and server processing.

## Affected methods
- find
- summarize
- text_search (API query_display_name_cache)

## Detailed Description

```python
sg.find('Asset', [['project', 'is', {'created_at': datetime.datetime(2015, 12, 16, 11, 2, 10, tzinfo), 'id': 72, 'name': 'Demo: Game', 'type': 'Project'}]]) 
```
```python
sg.find('Asset', [['project', 'is', {'id': 72, 'type': 'Project'}]]) 
```

This version also includes a new environmental variable to enable this feature. Please use it **_at initialization_** if you want to compare results or if you are getting any unexpected behavior.

```bash
export SHOTGUN_API_ENABLE_ENTITY_OPTIMIZATION =1
```

## Testing
Install this `v3.7.0-beta.1` version.

```sh
pip install git+https://github.com/shotgunsoftware/python-api.git@v3.7.0-beta.1
```

In my first tests, I added 40 fake tests to a project dict and with this feature applied, the result time decreased from 0.9726s to 0.3405s.

```python
os.environ["SHOTGUN_API_ENABLE_ENTITY_OPTIMIZATION"] = "1"
sg = shotgun_api3.Shotgun(...

# Getting project fields
project_fields = list(sg.schema_field_read("Project").keys())

# Getting a project
sg_full_project = sg.find_one('Project', [['id', 'is', PROJECT_ID]], project_fields)

# At this point, `sg_full_project` can be "hacked" with more unnecessary fields to simulate a larger payload.

# Getting an asset
sg_asset = sg.find('Asset', [['project', 'is', sg_full_project]])
```

## TODO

- [x] Publish `v3.7.0-beta.1` tag for testing


shotgun_api3/shotgun.py
```diff
@@ -4470,6 +4470,28 @@ def _translate_filters_simple(sg_filter):
     if len(values) == 1 and isinstance(values[0], (list, tuple)):
         values = values[0]
 
+    # Payload optimization: Do not send a full object
+    # just send the `type` and `id` when combining related queries
+    disable_entity_optimization = os.environ.get(
+        "SHOTGUN_API_DISABLE_ENTITY_OPTIMIZATION"
```
eduardoChaucaGallegos, line 4476: Wouldn't it be better if it were dynamic? For instance, if the query only includes the `id` and `type `fields, it could be made with just that data, without relying on the environment variable. Because what happens if I forget to deactivate the environment variable and use more parameters than just `id` and `type`?

carlos-villavicencio-adsk: The point of having this environmental variable is only to "disable" this new feature and behave as it was. It can help ups detect any potential issues while leaving a quick fix for the testers.

shotgun_api3/shotgun.py
```diff
@@ -4470,6 +4470,28 @@ def _translate_filters_simple(sg_filter):
...
+    )
+    if (
+        not disable_entity_optimization
```
julien-lang, line 4479: One problem with this logic, this will match if `SHOTGUN_API_DISABLE_ENTITY_OPTIMIZATION=1`

There is no behavior difference between values 0 and 1 which seems pretty bad.

I'd prefer to convert the value from string to bool 
```python
disable_entity_optimization = os.environ.get(
  "SHOTGUN_API_DISABLE_ENTITY_OPTIMIZATION", ""
).strip().lower() not in ["0", "no", "false"]
```

shotgun_api3/shotgun.py
```diff
@@ -4470,6 +4470,28 @@ def _translate_filters_simple(sg_filter):
...
+        and condition["path"] != "id"
+        and condition["relation"] in ["is", "is_not"]
+        and isinstance(values[0], dict)
+    ):
+        try:
+            values = [
+                {
+                    "type": values[0]["type"],
+                    "id": values[0]["id"],
+                }
```
julien-lang, line 4489: Why do we keep a dict?

I see in the description that we are reducing the payload to:
```python
sg.find('Asset', [['project', 'is', {'id': 72, 'type': 'Project'}]])
```

But why not to:
```python
sg.find('Asset', [['project', 'is', 72}]])
```

instead?

shotgun_api3/shotgun.py
```diff
@@ -4470,6 +4470,28 @@ def _translate_filters_simple(sg_filter):
...
+            ]
+        except KeyError:
+            pass
+
+
```
julien-lang, line 4494: No 2 empty lines in a function's body

docs/reference.rst
```diff
@@ -949,6 +949,30 @@ Stores the number of milliseconds to wait between request retries.  By default,
...
+=======================================
+
+When set to ``1``, this environment variable will disable the entity optimization feature.  This feature is enabled by default and is used to reduce the payload size made to the server when retrieving entities improving performance.  
+
+For example, a ``find`` call like this:
+
+.. code-block:: python
+
+    # sg_full_project is a project entity dictionary with a lot of FPTR fields
+    sg.find('Asset', [['project', 'is', sg_full_project]])
```
julien-lang, line 963: What you mean by sg_full_project is not clear here. We need a more detailed example I think. Like the one in the PR description for instance.

docs/reference.rst
```diff
@@ -949,6 +949,30 @@ Stores the number of milliseconds to wait between request retries.  By default,
...
+
+
+Will internally be transformed as if you invoked something like this:
+
+.. code-block:: python
+
+    sg.find('Asset', [['project', 'is', {'id': 999, 'type': 'Project'}]]) 
+
+
+When disabled, the API will make a request with the original entity data in the ``fields`` argument. Feel free to disable this feature if you are experiencing issues with the entity optimization feature.
```
julien-lang, line 973: Which was the default and only behavior before version 3.7

shotgun_api3/shotgun.py
carlos-villavicencio-adsk: That won't be accepted by SG

shotgun_api3/shotgun.py
carlos-villavicencio-adsk: Thanks. I was aware of that and I this it's a valid option for testing purposes.

shotgun_api3/shotgun.py
carlos-villavicencio-adsk: I realized we don't have code linting here.

docs/reference.rst
carlos-villavicencio-adsk: > sg_full_project is a project entity dictionary with a lot of FPTR fields

In other words, it's the result from 

```python
sg.find_one('Project', [['id', 'is', PROJECT_ID]], fields)
```

I can add that in the docs anyway

shotgun_api3/shotgun.py
julien-lang: Well I don't understand. This is exactly like this isn't it?
image

carlos-villavicencio-adsk: It's not. Check the first argument of `find` or `find_one`. When entity `Project` is sent, you can pass directly the id as you want. When doing a relate query by sending an `Asset` entity, AND you want to query but asset's project ID, then you need to provide a project object dict.

docs/reference.rst
```diff
@@ -949,6 +949,30 @@ Stores the number of milliseconds to wait between request retries.  By default,
...
+SHOTGUN_API_DISABLE_ENTITY_OPTIMIZATION
+=======================================
+
+When set to ``1``, this environment variable will disable the entity optimization feature.  This feature is enabled by default and is used to reduce the payload size made to the server when retrieving entities improving performance.  
+
+For example, a ``find`` call like this:
+
+.. code-block:: python
+
+    sg.find('Asset', [['project', 'is', {'created_at': datetime.datetime(2015, 12, 16, 11, 2, 10, tzinfo), 'id': 9999, 'name': 'Demo: Game', 'type': 'Project', ...}]])
```
julien-lang, line 962: Could you add some line split in this code please so it's readable without horizontal scrolling

docs/reference.rst
```diff
@@ -949,6 +949,35 @@ Stores the number of milliseconds to wait between request retries.  By default,
 
 In the case that both this environment variable and the config's ``rpc_attempt_interval`` property are set, the value in ``rpc_attempt_interal`` will be used.
 
+
+SHOTGUN_API_DISABLE_ENTITY_OPTIMIZATION
+=======================================
```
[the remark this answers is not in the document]
carlos-villavicencio-adsk: I see your point and I will change it to set the env var to enable the feature. For now (testing phase).

However, in the future the idea is to change the behavior to be the default. In that case, I think we still need to provide a way to disable this feature (mostly for debugging purposes).

eduardoChaucaGallegos (approved):

shotgun_api3/shotgun.py
```diff
@@ -4470,6 +4470,27 @@ def _translate_filters_simple(sg_filter):
     if len(values) == 1 and isinstance(values[0], (list, tuple)):
         values = values[0]
 
+    # Payload optimization: Do not send a full object
+    # just send the `type` and `id` when combining related queries
+    enable_entity_optimization = os.environ.get(
+        "SHOTGUN_API_ENABLE_ENTITY_OPTIMIZATION", "0"
+    ).strip().lower()
+    if (
+        enable_entity_optimization not in ["0", "no", "false"]
```
julien-lang, line 4479: Minor point but I'd prefer:
```suggestion
    enable_entity_optimization = os.environ.get(
        "SHOTGUN_API_ENABLE_ENTITY_OPTIMIZATION", "0"
    ).strip().lower() in ["1", "yes", "true"]
    if (
        enable_entity_optimization
```

Because:

1. Thinking about it, I prefer that we set to true when the setting value is clearly indicates a "true" value (instead of interpreting "is not false")
2. The first line goal becomes more clear as "read the preference and obtain a boolean value" while the second line is pure algorithm "if true then"
  * In the current code, I feel that the second line (if) does 2 different things: 1. convert the pref to boolean and 2. if true then enable...




Then I'd move the comment above (what this feature does) in the IF block (first lines of the block). But this might only be a personal preference

julien-lang (commented):
Another comment. Although not a blocker.

shotgun_api3/shotgun.py
```diff
@@ -4470,6 +4470,27 @@ def _translate_filters_simple(sg_filter):
     if len(values) == 1 and isinstance(values[0], (list, tuple)):
         values = values[0]
 
+    # Payload optimization: Do not send a full object
+    # just send the `type` and `id` when combining related queries
+    enable_entity_optimization_value = os.environ.get(
+        "SHOTGUN_API_ENABLE_ENTITY_OPTIMIZATION", "0"
+    ).strip().lower()
+    if (
+        enable_entity_optimization_value == "1"
```
julien-lang, line 4479: Another round of suggestions on this block:

```suggestion
    use_entity_optimization = os.environ.get(
        "SHOTGUN_API_ENABLE_ENTITY_OPTIMIZATION", "0"
    ).strip().lower() in ["1", "yes", "true"]

    if (
        use_entity_optimization == "1"
```

This mostly comes from my [previous comment](https://github.com/shotgunsoftware/python-api/pull/360#discussion_r1851141819).

julien-lang (commented):
Could you update the env var name in the description too please? 

It's still says:
```bash
export SHOTGUN_API_DISABLE_ENTITY_OPTIMIZATION=1
```

Instead of:
```bash
export SHOTGUN_API_ENABLE_ENTITY_OPTIMIZATION=1
```
Looking at the description some more, given that the optimization is  disabled by default

shotgun_api3/shotgun.py
carlos-villavicencio-adsk: Let's enforce a single feature flag value of `1`. The point is to increase performance so I'm dropping a list lookup that is O(n) in favor of a single comparison that I guess is O(1).

julien-lang: Thinking some more about it, I would prefer that the setting definition be in the "Config" object so it's only interpreted once:

```diff
diff --git a/shotgun_api3/shotgun.py b/shotgun_api3/shotgun.py
index 99b2a59..bfbe06e 100644
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -404,6 +404,10 @@ class _Config(object):
         # In the case that the environment variable is already set, setting the
         # property on the config will override it.
         self.rpc_attempt_interval = 3000
+
+        # TODO
+        self.use_payload_optmiziation = False
+
         # From http://docs.python.org/2.6/library/httplib.html:
         # If the optional timeout parameter is given, blocking operations
         # (like connection attempts) will timeout after that many seconds
@@ -650,6 +654,22 @@ class Shotgun(object):
             raise ValueError("Value of SHOTGUN_API_RETRY_INTERVAL must be positive, "
                              "got '%s'." % self.config.rpc_attempt_interval)
 
+        try:
+            use_payload_optmiziation = os.environ.get(
+                "SHOTGUN_API_ENABLE_ENTITY_OPTIMIZATION"
+            ).strip().lower()
+            if use_payload_optmiziation in ["1", "yes", "true"]:
+                self.config.use_payload_optmiziation = True
+            elif use_payload_optmiziation in ["0", "no", "false"]:
+                self.config.use_payload_optmiziation = False
+            else:
+                LOG.debug(
+                    f'Unrecognized value "{use_payload_optmiziation}" for SHOTGUN_API_ENABLE_ENTITY_OPTIMIZATION.'
+                    ' Sticking to default value ({self.config.use_payload_optmiziation})'
+                )
+        except KeyError:
+            pass
+
         self._connection = None
 
         self.__ca_certs = self._get_certs_file(ca_certs)

```

carlos-villavicencio-adsk: I refactored the code so that the env var is now only read in the `Shotgun` constructor. I didn't use the config object because the filtering is done in functions outside the `Shotgun` class.

I think moving these functions to the `Shotgun` class is risky for now.

julien-lang (dismissed):
I don't like using `global` but we already use some there I'll let it go.

PR #362, opened 2024-12-19
SG-37544 Include the "in" and "not_in" operators for payload optimization

carlos-villavicencio-adsk:
## Summary

This is a follow up work from #360 in this case including support for `in` and `not_in` [operators](https://developers.shotgridsoftware.com/python-api/reference.html#operators-and-arguments) for lists of entities.

## Affected methods
- find
- summarize
- text_search (API query_display_name_cache)

## Detailed Description

```python
sg.find("Asset",
        [[
            "project", "in",
            [
                {
                    "created_at": datetime.datetime(2015, 12, 16, 11, 2, 10, tzinfo),
                    "id": 72,
                    "name": "Demo: Game",
                    "type": "Project",
                    # ...
                },
                {
                    "created_at": datetime.datetime(2015, 12, 16, 11, 2, 10, tzinfo),
                    "id": 73,
                    "name": "Demo: Game 2",
                    "type": "Project",
                    # ...
                }
            ]
        ]]
)

```
```python
sg.find('Asset', [['project', 'in', [{'id': 72, 'type': 'Project'}, {'id': 73, 'type': 'Project'}]]]) 
```

### DIFF
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -12,6 +12,7 @@
 
 import os
 import unittest
+from unittest import mock
 from .mock import patch
 import shotgun_api3 as api
 from shotgun_api3.shotgun import _is_mimetypes_broken
@@ -434,6 +435,7 @@ class TestFilters(unittest.TestCase):
         result = api.shotgun._translate_filters(filters, "all")
         self.assertEqual(result, expected)
 
+    @mock.patch.dict(os.environ, {"SHOTGUN_API_ENABLE_ENTITY_OPTIMIZATION": "1"})
     def test_related_object_entity_optimization_is(self):
         filters = [
             [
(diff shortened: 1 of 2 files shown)


tests/test_unit.py
```diff
@@ -462,6 +462,41 @@ def test_related_object_entity_optimization(self):
...
+                        },
+                        {
+                            "id": 998,
+                            "type": "Anything",
+                        }
+                    ],
+                }
+            ],
+        }
+        os.environ["SHOTGUN_API_ENABLE_ENTITY_OPTIMIZATION"] = "1"
```
julien-lang, line 495: For this kind of test, in order to leave in the same state, I prefer to use a context:
```python
from unittest import mock
with mock.patch.dict("os.environ", {
  "SHOTGUN_API_ENABLE_ENTITY_OPTIMIZATION": "1",
}):
```

carlos-villavicencio-adsk: I've found a way using a decorator for the test case.

julien-lang (approved):
LGTM!

eduardoChaucaGallegos (approved):

PR #363, opened 2024-12-20
SG-37548 Add payload optimization on update method

carlos-villavicencio-adsk:
## Summary

This is a follow up work from #360 in this case including support for `update` [method](https://developers.shotgridsoftware.com/python-api/reference.html#shotgun_api3.shotgun.Shotgun.update).

## Affected methods
- update

## Detailed Description

```python
sg.update(
    "Asset", 123,
    {
        "name": "Some name",
        "project": {
            "created_at": datetime.datetime(2015, 12, 16, 11, 2, 10, tzinfo),
            "id": 72,
            "name": "Demo: Game",
            "type": "Project",
        },
    },
)

```
```python
sg.update('Asset', 123, {'name': 'Some name', 'project' : {'id': 72, 'type': 'Project'}})
```

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -4526,6 +4526,5 @@ def _get_type_and_id_from_value(value):
             return [{"type": v["type"], "id": v["id"]} for v in value]    
     except (KeyError, TypeError):
         LOG.debug(f"Could not optimize entity value {value}")
-        pass
 
     return value


shotgun_api3/shotgun.py
```diff
@@ -1132,6 +1132,31 @@ def _add_project_param(self, params, project_entity):
...
 
         return params
+    
+    def _translate_update_params(
+        self, entity_type, entity_id, data, multi_entity_update_modes
+    ):
+        global SHOTGUN_API_ENABLE_ENTITY_OPTIMIZATION
+
+        def optimize_value(field_value):
+            if isinstance(field_value, dict) and SHOTGUN_API_ENABLE_ENTITY_OPTIMIZATION:
```
eduardoChaucaGallegos, line 1142: Isn't it better?
```
if SHOTGUN_API_ENABLE_ENTITY_OPTIMIZATION and isinstance(field_value, dict): 
   ...
```
cause if SHOTGUN_API_ENABLE_ENTITY_OPTIMIZATION is normally False, this form might perform an unnecessary evaluation of isinstance(field_value, dict).

carlos-villavicencio-adsk: Good call. I'll do that suggestion

eduardoChaucaGallegos (approved):

julien-lang (dismissed):

shotgun_api3/shotgun.py
```diff
@@ -4500,3 +4512,19 @@ def _version_str(version):
...
+    """
+    try:
+        if isinstance(value, dict):
+            return {"type": value["type"], "id": value["id"]}
+        elif isinstance(value, list):
+            return [{"type": v["type"], "id": v["id"]} for v in value]
+        else:
+            return value
+    except (KeyError, TypeError):
+        return value
```
julien-lang, line 4530:
```suggestion
def _get_type_and_id_from_value(value):
    """
    For an entity dictionary, returns a new dictionary with only the type and id keys.
    If any of these keys are not present, the original dictionary is returned.
    """
    try:
        if isinstance(value, dict):
            return {"type": value["type"], "id": value["id"]}
        elif isinstance(value, list):
            return [{"type": v["type"], "id": v["id"]} for v in value]
    except (KeyError, TypeError):
        pass

    return value
```

julien-lang, line 4530: We might want a debug log here.

tests/test_unit.py
```diff
@@ -13,6 +13,7 @@
 import os
 import unittest
 from unittest import mock
+from unittest import mock
```
julien-lang, line 16: ?? Sounds useless

shotgun_api3/shotgun.py
```diff
@@ -4523,8 +4523,9 @@ def _get_type_and_id_from_value(value):
...
             return {"type": value["type"], "id": value["id"]}
         elif isinstance(value, list):
-            return [{"type": v["type"], "id": v["id"]} for v in value]
-        else:
-            return value
+            return [{"type": v["type"], "id": v["id"]} for v in value]    
     except (KeyError, TypeError):
-        return value
+        LOG.debug(f"Could not optimize entity value {value}")
+        pass
```
julien-lang, line 4529: No need for `pass` anymore if you have an instruction in the block
```suggestion
        LOG.debug(f"Could not optimize entity value {value}")
```

julien-lang (approved):

tests/test_unit.py
```diff
@@ -290,6 +290,8 @@ def test_py_version(self, mock_sys):
 
 
 class TestFilters(unittest.TestCase):
+    maxDiff = None
```
eduardoChaucaGallegos, line 293: Do you know if we use this variable? why do we use camelcase here?

carlos-villavicencio-adsk: Yes, [maxDiff](https://docs.python.org/3.11/library/unittest.html#unittest.TestCase.maxDiff) is used to show the full difference of the assert string in case of failure. I enable this at full difference (setting to `None`) because this test suite handles medium-sized strings.

eduardoChaucaGallegos (approved):

PR #368, opened 2025-02-11
SG-35712 Prevent flaky disconnection when uploading thumbnails on publish

carlos-villavicencio-adsk:
There's a flaky disconnection when the publisher uploads the thumbnail. The most common errors are:

- Connection closed by peer
- URLopen error EOF occurred in violation of protocol ssl.c:1006

They seem to be related to TLS handshakes. While testing on local machines we consistently reproduced this with the following change: Move the connection opener to the retry block and change `CACertsHTTPSHandler` parent class.

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -4293,7 +4293,7 @@ class Shotgun(object):
             raise ShotgunError("Max attemps limit reached.")
 
 
-class CACertsHTTPSConnection(http_client.HTTPConnection):
+class CACertsHTTPSConnection(http_client.HTTPSConnection):
     """"
     This class allows to create an HTTPS connection that uses the custom certificates
     passed in.
@@ -4309,11 +4309,11 @@ class CACertsHTTPSConnection(http_client.HTTPConnection):
         """
         # Pop that argument,
         self.__ca_certs = kwargs.pop("ca_certs")
-        http_client.HTTPConnection.__init__(self, *args, **kwargs)
+        super().__init__(self, *args, **kwargs)
 
     def connect(self):
         "Connect to a host on a given (SSL) port."
-        http_client.HTTPConnection.connect(self)
+        super().connect(self)
         # Now that the regular HTTP socket has been created, wrap it with our SSL certs.
         if six.PY38:
             context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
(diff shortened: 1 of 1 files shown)


shotgun_api3/shotgun.py
```diff
@@ -4309,19 +4309,19 @@ def __init__(self, *args, **kwargs):
...
+        super().connect(self)
         # Now that the regular HTTP socket has been created, wrap it with our SSL certs.
         if six.PY38:
             context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
             context.verify_mode = ssl.CERT_REQUIRED
             context.check_hostname = False
             if self.__ca_certs:
                 context.load_verify_locations(self.__ca_certs)
-            self.sock = context.wrap_socket(self.sock)
+            self.sock = context.wrap_socket(self.sock, server_hostname=self.host)
```
julien-lang, line 4324: Why do we need this change?

shotgun_api3/shotgun.py
```diff
@@ -4330,13 +4330,13 @@ def connect(self):
             )
 
 
-class CACertsHTTPSHandler(urllib.request.HTTPSHandler):
+class CACertsHTTPSHandler(urllib.request.HTTPHandler):
```
julien-lang, line 4333: Why do we need this change?

shotgun_api3/shotgun.py
```diff
@@ -4267,12 +4267,12 @@ def _send_form(self, url, params):
 
         params.update(self._auth_params())
 
-        opener = self._build_opener(FormPostHandler)
 
         attempt = 1
         while attempt <= self.MAX_ATTEMPTS:
             # Perform the request
             try:
+                opener = self._build_opener(FormPostHandler)
```
julien-lang, line 4275: Side question: why don't we use httplib2 for those?

carlos-villavicencio-adsk: I'm not sure how to answer this. This line was moved from outside the retry block, so in case it fails, the retry can start a new connection. I don't know the details of the original design.

shotgun_api3/shotgun.py
carlos-villavicencio-adsk: I don't have a clear answer for this. I was a trial/error change that made sense to me according to keep consistency on naming. Also, we're using SSL with TLS.

shotgun_api3/shotgun.py
carlos-villavicencio-adsk: It was not needed. Removed.

eduardoChaucaGallegos (approved):

julien-lang (approved):

PR #379, opened 2025-05-05
SG-38213 Prevent unexpected retries on error

carlos-villavicencio-adsk:
We want to control when python-api does a retry on error. 
- `SSLEOFError`
- `SSLHandshakeError`

On `_make_call` function. 

We don't want to retry on a general exceptions (e.g. Timeout or remote disconnection) because we might send a resource modification request (create, batch create, etc) and we can end up duplicating things. It seems this concern was already [identified before](https://github.com/shotgunsoftware/python-api/pull/196/files#r289380731).


shotgun_api3/shotgun.py
```diff
@@ -3938,11 +3938,10 @@ def _make_call(self, verb, path, body, headers):
...
                     LOG.debug("Request failed.  Giving up after %d attempts." % attempt)
                     raise
-            except Exception:
+            except Exception as e:
                 self._close_connection()
-                if attempt == max_rpc_attempts:
-                    LOG.debug("Request failed.  Giving up after %d attempts." % attempt)
-                    raise
+                LOG.debug(f"Request failed.  Reason: {e}")
+                raise
```
julien-lang, line 3944: Why do we even keep this except block then?
Just for the `_close_connection`?

julien-lang (commented):
Looks good but I am worried we are going to create new regressions with removing that retry.

Did you identify when and why this retry on generic exceptions was introduced?

eduardoChaucaGallegos (dismissed):

shotgun_api3/shotgun.py
julien-lang: I don't understand this answer.

shotgun_api3/shotgun.py
```diff
@@ -3938,11 +3938,10 @@ def _make_call(self, verb, path, body, headers):
                 if attempt == max_rpc_attempts:
                     LOG.debug("Request failed.  Giving up after %d attempts." % attempt)
                     raise
-            except Exception:
+            except Exception as e:
                 self._close_connection()
-                if attempt == max_rpc_attempts:
-                    LOG.debug("Request failed.  Giving up after %d attempts." % attempt)
-                    raise
+                LOG.debug(f"Request failed.  Reason: {e}")
```
julien-lang, line 3943: Let's get the traceback since it's a debug log so we'll get more context

```suggestion
                LOG.debug(f"Request failed.  Reason: {e}", exc_info=True)
```

julien-lang (commented):
Just 2 minor points, then I'll be happy :)

shotgun_api3/shotgun.py
carlos-villavicencio-adsk: Ok let me summarize it: We use this except block for timeout or any other server error. And this is what's been happening with this ticket's context.

julien-lang (approved):
LGTM!

julien-lang (approved):
LGTM!

Just one minor comment

julien-lang (approved):

Issue #233, opened 2020-07-12
Param type incorrectly set to int instead of list in the docstring.


markthielen-dvs:
https://github.com/shotgunsoftware/python-api/blob/b3034c1901b13914e7a6f5eb6b7c938d51d061e1/shotgun_api3/shotgun.py#L851

PR #393, opened 2025-06-18
SG-39414 Add type annotations

### LINKED ISSUE #307: Add Type hints to common API methods
Now that shotgun is supposed to by py3.7 only, it would be great to add type hints for a better experience in IDE. currently in VS code with Pylance for instance, the inferred types for `find()` aren't great:
```python
def find(
    self: Self@Shotgun,
    entity_type: Unknown,
    filters: Unknown,
    fields: Unknown | None = None,
    order: Unknown | None = None,
    filter_operator: Unknown | None = None,
    limit: int = 0,
    retired_only: bool = False,
    page: int = 0,
    include_archived_projects: bool = True,
    additional_filter_presets: Unknown | None = None
) -> (list[Unknown] | tuple[Unknown])
```

If we explicitly type hint it like so:
```python
    def find(
        self,
        entity_type: str,
        filters: Iterable[Union[Iterable, Dict[str, Any]]],
        fields: Optional[Iterable[str]] = None,
        order: Optional[Iterable[Dict[str, str]]] = None,
        filter_operator: Union[Literal['all'], Literal['any']] = None,
        limit: int = 0,
        retired_only: bool = False,
        page: int = 0,
        include_archived_projects: bool = True,
        additional_filter_presets: Optional[Iterable[str]]= None,
    ) -> List[Dict[str, Any]]:
```

We would have a better experience with static type checkers, and a better experience interacting with the shotgun API from our IDEs.

If for some reason we still want to support python 2 despite officially being 3.7+, we can do the same thing with python 2 compatible comment-style type hints.

Either way, adding hints to the most common methods (find, find_one, create, update, delete, revive, batch, schema_read, ...) would make development smoother!

chadrik:
Resolves #307 and #233

When running `mypy` against the code there are still some errors, but the annotations should provide some good value to users without any real downsides.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ setup(
     packages=find_packages(exclude=("tests",)),
     script_args=sys.argv[1:],
     include_package_data=True,
-    package_data={"": ["cacerts.txt", "cacert.pem"]},
+    package_data={"": ["cacerts.txt", "cacert.pem", "py.typed"]},
     zip_safe=False,
     python_requires=">=3.9.0",
     classifiers=[


shotgun_api3/shotgun.py
```diff
@@ -896,15 +909,15 @@ def info(self):
...
-    ):
+        entity_type: str,
+        filters: list,
+        fields: Optional[list] = None,
+        order: Optional[list] = None,
+        filter_operator: Optional[Literal["all", "any"]] = None,
+        retired_only: bool = False,
+        include_archived_projects: bool = True,
+        additional_filter_presets: Optional[list] = None,
+    ) -> Optional[dict[str, Any]]:
```
austinwitherspoon, line 920: Some of these are not as complete as what you added to `find()`, for example `fields` here is `Optional[list]` but in `find()` it's `Optional[list[str]]`

I think what you're doing in `find()` is preferable and you should copy it here!

austinwitherspoon (commented):
It would be cool to add a blank `py.typed` marker file so tools like mypy know that this package has type hints!

https://peps.python.org/pep-0561/#packaging-type-information

I believe we'd also need to update setup.py to include that in the package data.

shotgun_api3/shotgun.py
chadrik: will do!

shotgun_api3/shotgun.py
```diff
@@ -76,6 +89,22 @@
...
+
+class GroupingItem(TypedDict):
+    field: str
+    type: str
+    direction: str
+
+
+class BaseEntity(TypedDict, total=False):
+    id: int
+    type: str
```
chadrik, line 105: I added some `TypedDicts` to try to fill in some of the required dictionary fields.  Let me know if you see anything wrong here.

shotgun_api3/shotgun.py
```diff
@@ -1450,7 +1486,12 @@ def summarize(
         records = self._call_rpc("summarize", params)
         return records
 
-    def create(self, entity_type, data, return_fields=None):
+    def create(
+        self,
+        entity_type: str,
+        data: dict[str, Any],
+        return_fields: Optional[list] = None,
+    ) -> dict[str, Any]:
```
chadrik, line 1494: This should be changed to the `BaseEntity`  `TypedDict`

chadrik:
This is a pretty low risk change, it would be great to get it merged.

chadrik:
Since this isn't getting any traction I'm going to push a stubs only library to pypi.   When this is complete I will remove it.

chadrik:
> I'd also be very excited to get these changes in! We currently maintain our own stubs package and it would be great to be able to delete it.

My stubs have been published to pypi, in case anyone needs them: https://pypi.org/project/types-shotgun-api3/

cdombrova-scanlinevfx:
Can this PR get some attention please?

PR #395, opened 2025-07-08
SG 38306 Python2 removal six and ensure_ functions

eduardoChaucaGallegos:
### **Description** 
- Removed six and ensure_ functions


shotgun_api3/shotgun.py
```diff
@@ -4684,7 +4635,7 @@ def connect(self):
         "Connect to a host on a given (SSL) port."
         super().connect(self)
         # Now that the regular HTTP socket has been created, wrap it with our SSL certs.
-        if six.PY38:
+        if sys.version_info >= (3, 8):
```
carlos-villavicencio-adsk, line 4638: We don't support 3.7 anymore, so I think we can get rid of this condition and leave only the truth block.

carlos-villavicencio-adsk (commented):
Just a small proposal.

shotgun_api3/lib/sgutils.py
julien-lang, line 1: That's only possible if shotgunsoftware/tk-core#1042 is merged and released first

carlos-villavicencio-adsk (approved):

PR #400, opened 2025-07-16
SG-38306 Python2 Removal - Part 3 - Cleanup imports

julien-lang:
* Replace `six.moves` imports
* Cleanup BytesIO import from six
* Remove deprecated Simple json import management
* Cleanup Py2-3 compat with ImportError
* Simplify Base64

# PR orders

* #398
* #399 
* #400 
* #401
* #402
* #403
* #404
* #405
* #395
* #372

PR #372, opened 2025-02-25
SG-38306 Remove Python 2 - Part 10 - SSL

julien-lang:
This PR removes SSL code handling edge cases that were only accurate in Python 2.
This is part of a series of PRs modernizing the codebase for Python 3 only.

- Removes the global `NO_SSL_VALIDATION` flag and related SSL validation fallback logic
- Simplifies HTTPS handler to use modern Python 3 SSL context creation
- Updates user agent strings to remove SSL validation status indicators

# PR orders

* #398
* #399 
* #400 
* #401
* #402
* #403
* #404
* #405
* #395
* #372

PR #401, opened 2025-07-16
SG-38306 Python2 Removal - Part 4 - Mimetypes module

julien-lang:
Remove now deprecated copy of Python 2.7's [mimetypes](https://docs.python.org/3/library/mimetypes.html) module. 
We prefer to use [Python's default mimetypes](https://docs.python.org/3/library/mimetypes.html) module instead since we are not aware of any issues with this module anymore.

# PR orders

* #398
* #399 
* #400 
* #401
* #402
* #403
* #404
* #405
* #395
* #372


carlos-villavicencio-adsk (approved):
🚢

eduardoChaucaGallegos (dismissed):

PR #415, opened 2025-10-01
SG-40026 Fixup super method should not pass self as first argument

julien-lang:
This PR is an update of #394 since the PR could not be merged as-is due to recent changes in the repo.

The scope of this PR is exactly the same as #394: Fix misused super in `CACertsHTTPSConnection`.

Closes #394


carlos-villavicencio-adsk (approved):

julien-lang:
Replaced by #418

PR #447, opened 2026-05-25
SG-43389 - Add OpenAPI 3.0.3 spec for the Python API 

camiloleal-globant:
[SG-43389](https://jira.autodesk.com/browse/SG-43389)

#### What was done:
Added `docs/openapi/spec.yaml` an OpenAPI 3.0.3 description of the Flow Production Tracking
Python API (`shotgun_api3`). Content is sourced directly from the
[API Reference documentation](https://developers.shotgridsoftware.com/python-api/reference.html).

#### Test:
- Validated syntax with `npx @redocly/cli lint docs/openapi/spec.yaml`. 0 errors
- Rendered interactive docs with `npx @redocly/cli build-docs docs/openapi/spec.yaml` to verify output
- Alternatively, paste the file into [editor.swagger.io](https://editor.swagger.io) or [redocly.com/editor](https://redocly.com/editor) for a visual review

image


docs/openapi/spec.yaml
```diff
@@ -0,0 +1,2103 @@
+openapi: "3.0.3"
+
+info:
+  title: "Flow Production Tracking Python API"
```
julien-lang: I think the real name is "Flow Production Tracking API3" or FlowPT API3. 
Python API is just the name of this repository

docs/openapi/spec.yaml
```diff
@@ -0,0 +1,2103 @@
...
+  version: "3.10.0"
```
julien-lang: Ok does that mean we need to update that field every time we release a new version to this library?

Basically, is this file describing the client library or the API? If it's the API, it might not be the best place to put it.

camiloleal-globant: You right, updated the title to Flow Production Tracking API3 and version to 1.0 to keep it decoupled from the Python library.

camiloleal-globant:
Handled in: https://git.autodesk.com/shotgun/shotgun/pull/20346

PR #449, opened 2026-06-04
SG-43825 Improve pipeline stability

jay-roe:
Azure pipelines were occasionally failing for various reasons. Some were transient connection errors (timeouts), but sometimes it was live state issues or test run pollution: only certain runners were failing.

This PR creates ephemeral projects scoped to Build + OS + Python version, so concurrent jobs are fully isolated. These projects are deleted unconditionally on job end.

We also added retries on the tests for the transient error cases.

Also modified `test_modify_visibility`, which was failing unpredictably (likely due to race conditions on using the same existing projects across runners). Now, it uses the ephemeral project for modifying the visibility, and uses the existing project *only* as a control group (so this one should never have it's visibility changed and not affect concurrent executions of `test_modify_visibility`)

Tested multiple times to ensure no more failures due to those identified issues
image

### DIFF
--- a/azure-pipelines-templates/run-tests.yml
+++ b/azure-pipelines-templates/run-tests.yml
@@ -102,6 +102,8 @@ jobs:
           --durations=0 \
           --nunit-xml=test-results.xml \
           --verbose \
+          --reruns 1 \
+          --reruns-delay 2 \
     env:
       # Tell Pytest that we're running in a CI environment
       CI: 1
--- a/azure-pipelines-templates/run-tests.yml
+++ b/azure-pipelines-templates/run-tests.yml
@@ -123,20 +123,42 @@ jobs:
       # This will give a user name like 'something macOS 2.7'
       SG_HUMAN_NAME: $(python_api_human_name) ${{ parameters.os_name }} ${{ parameters.python_version }}
       SG_HUMAN_PASSWORD: $(python_api_human_password)
-      # So, first, we need to make sure that two builds running at the same time do not manipulate
-      # the same entities, so we're sandboxing build nodes based on their name.
-      SG_PROJECT_NAME: Python API CI - $(Agent.Name)
-      # The entities created and then reused between tests assume that the same user is always
-      # manipulating them. Because different builds will be assigned different agents and therefore
-      # different projects, it means each project needs to have an entity specific to a given user.
-      # Again, this would have been a lot simpler if we could simply have had a login based on the
-      # agent name, but alas, the agent name has a space in it which needs to be replaced to something
-      # else and string substitution can't be made on build variables, only template parameters.
-      SG_ASSET_CODE: CI-$(python_api_human_login)-${{ parameters.os_name }}-${{ parameters.python_version }}
-      SG_VERSION_CODE: CI-$(python_api_human_login)-${{ parameters.os_name }}-${{ parameters.python_version }}
-      SG_SHOT_CODE: CI-$(python_api_human_login)-${{ parameters.os_name }}-${{ parameters.python_version }}
-      SG_TASK_CONTENT: CI-$(python_api_human_login)-${{ parameters.os_name }}-${{ parameters.python_version }}
-      SG_PLAYLIST_CODE: CI-$(python_api_human_login)-${{ parameters.os_name }}-${{ parameters.python_version }}
+      # Each job gets its own ephemeral project, scoped to this build + OS + Python version.
+      # This eliminates state bleed between concurrent runs and across successive builds on the
+      # same agent. The project is retired in the "Cleanup test project" step below.
+      SG_PROJECT_NAME: Python API CI - $(Build.BuildId) - ${{ parameters.os_name }} - ${{ parameters.python_version }}
+      # Entity codes only need to be unique within the project, so fixed strings are fine.
+      SG_ASSET_CODE: CI-asset
+      SG_VERSION_CODE: CI-version
+      SG_SHOT_CODE: CI-shot
+      SG_TASK_CONTENT: CI-task
+      SG_PLAYLIST_CODE: CI-playlist
+
+  - task: Bash@3
+    displayName: Cleanup test project
+    condition: always()
+    inputs:
+      targetType: inline
+      script: |
+        python -c "
+        import os, shotgun_api3
+        sg = shotgun_api3.Shotgun(
+            os.environ['SG_SERVER_URL'],
+            os.environ['SG_SCRIPT_NAME'],
+            os.environ['SG_API_KEY'],
+        )
+        project = sg.find_one('Project', [['name', 'is', os.environ['SG_PROJECT_NAME']]])
+        if project:
+            sg.delete('Project', project['id'])
+            print('Retired project:', os.environ['SG_PROJECT_NAME'])
+        else:
+            print('Project not found, nothing to clean up.')
+        "
+    env:
+      SG_SERVER_URL: $(ci_site)
+      SG_SCRIPT_NAME: $(ci_site_script_name)
+      SG_API_KEY: $(ci_site_script_key)
+      SG_PROJECT_NAME: Python API CI - $(Build.BuildId) - ${{ parameters.os_name }} - ${{ parameters.python_version }}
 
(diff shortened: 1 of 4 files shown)


tests/base.py
```diff
@@ -289,7 +289,7 @@ def _setup_db(cls, config, sg):
         cls.human_user = _find_or_create_entity(sg, "HumanUser", data)
 
         data = {"code": cls.config.asset_code, "project": cls.project}
-        keys = ["code"]
+        keys = ["code", "project"]
```
jay-roe, line 292: Since we are now creating many projects with entities of the same name, we must scope the search to the ephemeral project.

azure-pipelines-templates/run-tests.yml
```diff
@@ -102,6 +102,8 @@ jobs:
           --durations=0 \
           --nunit-xml=test-results.xml \
           --verbose \
+          --reruns 1 \
+          --reruns-delay 2 \
```
julien-lang, line 106: Do we need that? I mean, if you did fix the bug, we should not right ;)

azure-pipelines-templates/run-tests.yml
```diff
@@ -123,20 +125,42 @@ jobs:
...
+      # Each job gets its own ephemeral project, scoped to this build + OS + Python version.
+      # This eliminates state bleed between concurrent runs and across successive builds on the
+      # same agent. The project is retired in the "Cleanup test project" step below.
+      SG_PROJECT_NAME: Python API CI - $(Build.BuildId) - ${{ parameters.os_name }} - ${{ parameters.python_version }}
+      # Entity codes only need to be unique within the project, so fixed strings are fine.
+      SG_ASSET_CODE: CI-asset
+      SG_VERSION_CODE: CI-version
+      SG_SHOT_CODE: CI-shot
+      SG_TASK_CONTENT: CI-task
+      SG_PLAYLIST_CODE: CI-playlist
```
julien-lang: Well, do we really need these variable now they all the same name (under dedicated projects)?

azure-pipelines-templates/run-tests.yml
```diff
@@ -123,20 +125,42 @@ jobs:
...
+
+  - task: Bash@3
+    displayName: Cleanup test project
+    condition: always()
+    inputs:
+      targetType: inline
+      script: |
+        python -c "
```
julien-lang: Can you run `task: PythonScript@0` instead?

tests/requirements.txt
```diff
@@ -11,3 +11,4 @@
 pytest
 pytest-cov
 pytest-nunit
+pytest-rerunfailures
```
julien-lang, line 14: Again, why do we need this?

julien-lang (commented):
Lookg pretty good! Thanks for the improvements.

Just a few comments.

azure-pipelines-templates/run-tests.yml
jay-roe: Yes I fixed the bug, but this one is for transient errors. 

Timeouts and other network errors sometimes popped up in my testing. There is no way to fix this type of error other than retrying the request.

azure-pipelines-templates/run-tests.yml
```diff
@@ -123,20 +125,42 @@ jobs:
...
+            os.environ['SG_SCRIPT_NAME'],
+            os.environ['SG_API_KEY'],
+        )
+        project = sg.find_one('Project', [['name', 'is', os.environ['SG_PROJECT_NAME']]])
+        if project:
+            sg.delete('Project', project['id'])
+            print('Retired project:', os.environ['SG_PROJECT_NAME'])
+        else:
+            print('Project not found, nothing to clean up.')
+        "
```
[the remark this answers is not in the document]
jay-roe: Nah it actually works, but to Julien's point, it's better to just use PythonScript for this

azure-pipelines-templates/run-tests.yml
```diff
@@ -123,20 +125,34 @@ jobs:
...
+    displayName: Cleanup test project
+    condition: always()
+    inputs:
+      scriptSource: inline
+      script: |
+        import os, shotgun_api3
+        sg = shotgun_api3.Shotgun(
+            os.environ['SG_SERVER_URL'],
+            os.environ['SG_SCRIPT_NAME'],
+            os.environ['SG_API_KEY'],
```
julien-lang: Can we use name parameter when available?

```suggestion
            script_name=os.environ['SG_SCRIPT_NAME'],
            api_key=os.environ['SG_API_KEY'],
```

azure-pipelines-templates/run-tests.yml
```diff
@@ -123,20 +125,34 @@ jobs:
...
+        )
+        project = sg.find_one('Project', [['name', 'is', os.environ['SG_PROJECT_NAME']]])
+        if project:
+            sg.delete('Project', project['id'])
+            print('Retired project:', os.environ['SG_PROJECT_NAME'])
+        else:
+            print('Project not found, nothing to clean up.')
```
julien-lang: Can we do "fail first"?

```suggestion
        if not project:
            print('Project not found, nothing to clean up.')
            sys.exit(0)

        sg.delete('Project', project['id'])
        print('Retired project:', os.environ['SG_PROJECT_NAME'])
```

julien-lang (commented):
Looks good. Just a last few minor comments.

azure-pipelines-templates/run-tests.yml
jay-roe: Good point. Removed this from the env vars. It will now default to whatever is in `tests/example_config`.

31ecccdd7c772732eb03afdf5c6324614b2ceff9

azure-pipelines-templates/run-tests.yml
jay-roe: 975d46d26467d19adbe75f673408245c76844128

tests/requirements.txt
jay-roe: Resolved above. Due to transient network errors

azure-pipelines-templates/run-tests.yml
jay-roe: 4d16f8da26222bc8c106852e43ef4e07b9e32ad6

azure-pipelines-templates/run-tests.yml
jay-roe: a3d814607e7f9a6668e1b9a6537fab24d5297816

azure-pipelines-templates/run-tests.yml
```diff
@@ -135,20 +135,25 @@ jobs:
...
     inputs:
       scriptSource: inline
+      workingDirectory: $(Build.SourcesDirectory)
       script: |
         import os, shotgun_api3
         sg = shotgun_api3.Shotgun(
-            os.environ['SG_SERVER_URL'],
-            os.environ['SG_SCRIPT_NAME'],
-            os.environ['SG_API_KEY'],
+            base_url=os.environ['SG_SERVER_URL'],
```
julien-lang: Sorry, no, not this one. base_url is a positional parameter, not a named parameter.

azure-pipelines-templates/run-tests.yml
```diff
@@ -123,20 +125,39 @@ jobs:
...
+      SG_PROJECT_NAME: Python API CI - $(Build.BuildId) - ${{ parameters.os_name }} - ${{ parameters.python_version }}
+
+  - task: PythonScript@0
+    displayName: Cleanup test project
+    condition: always()
+    inputs:
+      scriptSource: inline
+      workingDirectory: $(Build.SourcesDirectory)
+      script: |
+        import os, shotgun_api3
```
julien-lang:
```suggestion
        import os, shotgun_api3, sys
```

julien-lang (changes requested):
Almost there, just 2 last comments

azure-pipelines-templates/run-tests.yml
jay-roe: Based on this line it looks like a named (but not optional) parameter. https://github.com/shotgunsoftware/python-api/blob/aca9d94567de3fe00e96d9f39d70a8854c99d021/shotgun_api3/shotgun.py#L500

julien-lang (approved):

PR #456, opened 2026-08-03
SG-43999: add custom entity config features

fauverick:
This PR introduces 4 new functions that allow users to configure custom entity (previously can only be done through the UI):

custom_entity_read: read the config of a custom entity
custom_entity_enable: enable a custom entity with default configuration. Use custom_entity_configure afterwards to set the display name or feature flags.
custom_entity_configure: configure the display name/configuration flags of an enabled custom entity
custom_entity_disable: disable a custom entity. Will return an error if the custom entity currently has existing records. Require passing force=True to disable anyway.
Should only be merged when these methods are supported in FlowPT (once other PRs in SG-43999 are merged)

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -105,6 +105,16 @@ The documentation for all of the methods you'll need in your scripts lives in he
     Shotgun.schema
     Shotgun.entity_types
 
+.. rubric:: Custom Entity Configuration
+
+.. autosummary::
+    :nosignatures:
+
+    Shotgun.custom_entity_read
+    Shotgun.custom_entity_enable
+    Shotgun.custom_entity_configure
+    Shotgun.custom_entity_disable
+
 
 Connection & Authentication
 ===========================
@@ -191,6 +201,17 @@ Methods allow you to introspect and modify the Shotgun schema.
 .. automethod:: Shotgun.schema_read
 .. automethod:: Shotgun.schema
 .. automethod:: Shotgun.entity_types
+
+Custom Entity Configuration
+===========================
+
+Methods to read and configure Custom Entities at the site level. They require administrator
+privileges and a server running v8.88.0 or higher.
+
+.. automethod:: Shotgun.custom_entity_read
+.. automethod:: Shotgun.custom_entity_enable
+.. automethod:: Shotgun.custom_entity_configure
+.. automethod:: Shotgun.custom_entity_disable
 
 **********
 Exceptions
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -329,6 +329,14 @@ class ServerCapabilities(object):
             {"version": (3, 3, 0), "label": "return thumbnail URLs"}, False
         )
 
+    def ensure_custom_entity_config_support(self) -> None:
+        """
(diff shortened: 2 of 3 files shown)


shotgun_api3/shotgun.py
```diff
@@ -329,6 +329,14 @@ def ensure_return_image_urls_support(self) -> bool:
             {"version": (3, 3, 0), "label": "return thumbnail URLs"}, False
         )
 
+    def ensure_custom_entity_config_support(self) -> None:
+        """
+        Ensures server has support for the custom entity config API (read, enable, configure, disable), added in v8.88.0.
+        """
+        self._ensure_support(
+            {"version": (8, 88, 0), "label": "custom entity config API"}
+        )
```
fauverick, line 338: notes: this version should be updated with the correct version once the feature is rolled out

docs/reference.rst
```diff
@@ -105,6 +105,16 @@ The documentation for all of the methods you'll need in your scripts lives in he
...
 
+.. rubric:: Custom Entity Configuration
+
+.. autosummary::
+    :nosignatures:
+
+    Shotgun.custom_entity_read
+    Shotgun.custom_entity_enable
+    Shotgun.custom_entity_configure
+    Shotgun.custom_entity_disable
```
julien-lang, line 116: I don't see the point having enable and disable action as API.

Basically, we want a CRUD for these custom entities. I don't know if Create and Delete are part of the scope. So basically, that should give us only 2 API methods:

- custom_entity_read
- custom_entity_update (display_name, status/enable/disable, ...)

Now, we might want to keep the `custom_entity_enable` and `custom_entity_disable` Python method as shortcuts to call custom_entity_update but then there should not be anything about display name in there.

camusj: Hi @julien-lang 
I was assigned this job that Rick left unfinished.

Thanks for the feedback.
This was actually already discussed and settled during the REST API review (https://git.autodesk.com/shotgun/shotgun/pull/20770#pullrequestreview-7917964). 

They considered combining enable/disable/configure into a single update, but rejected it: the same fields (entity_config, force) mean different things depending on whether enabled is true or false, and default-value behavior differs between first-enable and a later update. 
Sooo a single endpoint would be more confusing, not less. That's why the server keeps activation and configuration separate.

What do you think about that?
Thanks again!

shotgun_api3/shotgun.py
```diff
@@ -3671,6 +3679,152 @@ def user_subscriptions_create(
...
+        :returns: The entity config snapshot dict with ``entity_type``, ``enabled``,
+            ``display_name``, and ``entity_config``.
+        :rtype: dict
+        :raises shotgun_api3.ShotgunError: if the entity type is invalid (fault code 104).
+        """
+        self.server_caps.ensure_custom_entity_config_support()
+
+        return self._call_rpc("custom_entity_read", {"entity_type": entity_type})
+
+    def custom_entity_enable(self, entity_type: str) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
camusj: Resolved by updating the PR description instead. `custom_entity_enable` intentionally only sets defaults. c`ustom_entity_configure` handles display name/flags. 
This matches the server contract (enable/configure are validated as separate operations).

PR #457, opened 2026-08-13
SG-44724 Fail fast on stale connections instead of hanging

jerrycheng9:
## Problem

`shotgun_api3` reuses one persistent HTTP(S) connection per `Shotgun` instance
across all RPC calls (`Shotgun._connection`, populated by `_get_connection()`;
the underlying socket is cached in `httplib2.Http.connections{}`), and every
request is sent with `Connection: keep-alive`.

If a NAT, firewall or load balancer along the path silently drops that idle TCP
session — no FIN, no RST — the client cannot tell. Only `TCP_NODELAY` was set on
the socket; there was no OS-level keepalive and no idle-age tracking on cached
connections. The next call writes into the send buffer successfully and then
blocks in `getresponse()` for the full `timeout_secs` before it can recover,
turning what should be a near-instant reconnect into a multi-hundred-second
hang.

Recovery does not help here, so the fix has to be preventive: `_make_call`
retries only `ssl.SSLEOFError` / `ssl.SSLError`, while its generic
`except Exception` closes the connection and re-raises with no retry, and
httplib2 re-raises `socket.timeout` unconditionally.

## Changes

Idle expiry (the actual fix):

- Add `_Config.max_connection_idle_secs`, default 60, which bounds how long a
  cached connection may sit idle before it is closed and recreated instead of
  reused. Tunable per instance; `None` or `0` disables expiry.
- Stamp `Shotgun._connection_last_used` in `_http_request` only *after* the
  request completes. A request that raised is no evidence the socket is alive,
  so it must not refresh the idle clock.
- `_get_connection()` closes and rebuilds a stale connection through the
  existing `_close_connection()`. Both connection consumers (`connect()` and
  `_http_request`) route through `_get_connection()`, so there is no bypass.

TCP keepalive (defense in depth):

- `KeepaliveHTTPConnection` / `KeepaliveHTTPSConnection` call
  `super().connect()` and then enable `SO_KEEPALIVE`, tuning idle/interval/count
  to 30/10/3 so the kernel can notice a vanished peer within roughly a minute of
  idling.
- Tuning is best effort: `SIO_KEEPALIVE_VALS` via ioctl on Windows,
  `TCP_KEEPIDLE`/`TCP_KEEPALIVE` elsewhere, skipping constants that do not exist
  on the platform and swallowing `OSError`. Bare `SO_KEEPALIVE` would not fire a
  probe for 7200s on default Linux settings, so the per-platform tuning is what
  makes this worth anything — but it is still secondary to the idle expiry.

### DIFF
--- a/shotgun_api3/shotgun.py
+++ b/shotgun_api3/shotgun.py
@@ -190,7 +190,26 @@ def _set_socket_keepalive(sock) -> None:
             LOG.debug("Unable to set %s on socket." % option_name, exc_info=True)
 
 
-class KeepaliveHTTPConnection(HTTPConnectionWithTimeout):
+class _KeepaliveConnectionMixin(http.client.HTTPConnection):
+    """
+    Mixin that enables TCP keepalive once the connection is established.
+
+    Must be listed before the httplib2 connection class so that this
+    ``connect()`` runs and delegates to the real one. ``self.sock`` is the
+    SSL-wrapped socket for HTTPS, which delegates ``setsockopt`` to the socket
+    underneath.
+
+    Derives from ``http.client.HTTPConnection``, the common base of both
+    httplib2 connection classes, so that ``super().connect()`` resolves for type
+    checkers. It is never instantiated on its own.
+    """
+
+    def connect(self) -> None:
+        super().connect()
+        _set_socket_keepalive(self.sock)
+
+
+class KeepaliveHTTPConnection(_KeepaliveConnectionMixin, HTTPConnectionWithTimeout):
     """
     httplib2 HTTP connection that enables TCP keepalive once connected.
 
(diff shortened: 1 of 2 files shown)


shotgun_api3/shotgun.py
```diff
@@ -123,6 +130,92 @@ class BaseEntity(TypedDict, total=False):
...
+        option = getattr(socket, option_name, None)
+        if option is None:
+            continue
+        try:
+            sock.setsockopt(socket.IPPROTO_TCP, option, value)
+        except OSError:
+            LOG.debug("Unable to set %s on socket." % option_name, exc_info=True)
+
+
+class KeepaliveHTTPConnection(HTTPConnectionWithTimeout):
```
dsauve-adsk: nit : Both KeepaliveHTTPConnection and KeepaliveHTTPSConnection are identical, maybe use a mixin to prevent duplication

dsauve-adsk (dismissed):
One minor comment

shotgun_api3/shotgun.py
```diff
@@ -4215,9 +4353,21 @@ def _inbound_visitor(value):
...
         """
         if self._connection is not None:
-            return self._connection
+            if self._is_connection_stale():
+                LOG.debug(
+                    "Connection has been idle for more than %s seconds, "
+                    "closing it and reconnecting."
+                    % self.config.max_connection_idle_secs
+                )
+                self._close_connection()
```
dsauve-adsk, line 4422: Small question, should that function  create a new connection when the connection is stale, not just close it?

jerrycheng9: _close_connection() resets self._connection to None. So it will fall through to build a new _connection below

dsauve-adsk (dismissed):
LGTM

dsauve-adsk (approved):
Still looks good.. Re-approving

PR #458, opened 2026-08-17
SG-44793  Remove deprecated run_appveyor script

julien-lang:
This is not used anymore and has been the case for multiple years.


jay-roe (approved):
Thanks for the trim ✂️

carlos-villavicencio-adsk (approved):

